PR #1, opened 2017-12-06
Add Getting-Started.md

bfranske:
Add some information on getting started and some basic testing that I was able to figure out. This should probably be mentioned in the Readme.md as well. It looks like the MQTT topic structure has changed from what is currently documented in the Readme. At least I was not able to use the insteon/state/AA.BB.CC and insteon/set/AA.BB.CC topics to control a device. Instead I needed to use as documented in this pull request like insteon/AA.BB.CC/set" -m '{ "state" : "on" }

### DIFF
--- a/Getting-Started.md
+++ b/Getting-Started.md
@@ -1,0 +1,28 @@
+## Getting Started with Insteon-MQTT in Debian Jessie
+
+### Installing Required Dependencies and Software
+1. Update your repositories: `apt update`
+1. Install Python 3 and pip: `apt install python3 python3-pip`
+1. Install required Python libraries: `pip3 install pyserial Jinja2 paho-mqtt pyyaml`
+1. Install Git (if pulling insteon-mqtt directly from github): `apt install git`
+1. Download latest copy of insteon-mqtt: `git clone https://github.com/TD22057/insteon-mqtt.git`
+1. Ensure you have an MQTT broker installed as well as `mosquitto_sub` and `mosquitto_pub` for testing.
+
+### Configure Insteon-MQTT config.yaml file
+- Set `port: "/dev/ttyUSB0"` or whatever the correct serial port is for your PLM
+- Set `address: xx.xx.xx` to the Insteon address of your PLM (printed on a label on the back of the PLM)
+- Set `startup_refresh: True` You may want to change this later if you have a large number of Insteon devices as it will create a bit of traffic on your Insteon network as each device is polled when you start insteon-mqtt but for now at least you probably want this enabled.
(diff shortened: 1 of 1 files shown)


TD22057:
Thanks - but give me a few days.  For consistency, I changed addresses to lower case and added topic control and payload templates with jinja2 in config.yaml.  I'm almost done with the command line tool to send updates into the server (basically it hides the MQTT sending of commands) at which point I'm going to do a complete update/rewrite of the documentation.  I should have a first cut (and first release) at it available this weekend.  At that point it would be great if you can try it out and suggest any improvements.

If you want to experiment with home assistant integration right now, check config.yaml which has instructions and the right topics and payloads for interacting with lights, dimmers, motion sensors, etc.

bfranske:
My particular interest is the MQTT/insteon bridge part. I actually use Node-RED and Javascript for my MQTT logic, not Home Assistant so the MQTT part would be the part I'm most interested in documenting.

You might also find some useful information about Insteon PLM and network commands in the Misterhouse project. https://github.com/hollie/misterhouse/tree/stable/lib It's probably the most complete Insteon protocol implementation I've seen thus far and is what I've been running so far to control my Insteon devices, I keep poking around looking for an MQTT bridge though because I'd much prefer to dump it and handle all my automation over MQTT.

TD22057:
Thanks - I used that quite a bit for some ideas.  https://github.com/pfrommerd/insteon-terminal is also a good resource for some of the db manipulation stuff.  FYI - I just pushed a big commit of the command line tool a few minutes ago.  I haven't finished testing all of the options but the basics are there.  I think the current config.yaml w/ the topic/payload template system should give you all the options you need for a custom MQTT set up but let me know if you see anything missing.  It also documents the MQTT API now - I'll be copying and expanding on that in the docs.

That completes support for all the devices I own (switches, dimmers, remotes, motion sensors, and smoke bridge).  I'm going to start re-writing all the docs later tonight.

FYI - once I get this more stable, phase 2 of the development will be support the misterhouse style scene system where the tool controls all of the all link db's on the devices and keeps them in sync with a configuration file.  I don't have very many devices at this point so that's not hugely important to me but I do plan on working on it at some point.  Once I have all the database modification commands tested, that type of work becomes pretty simple.  That will also allow the creation of custom PLM modem scenes and control of them via MQTT.

TD22057:
Thanks for the write up.  I prefer a virtualenv approach so I added a similar guide here:
https://github.com/TD22057/insteon-mqtt/blob/master/doc/quick_start.md
as well as a more complicated system install guide here:
https://github.com/TD22057/insteon-mqtt/blob/master/doc/auto_start.md

PR #35, opened 2018-02-11
KeypadLinc uses correct handler for scene commands

### LINKED ISSUE #34: KeypadLinc uses wrong handler for scene commands
KeypadLinc is using the `set` handler for `scene` commands. Pull request coming.

j3kestrel:
Fixes #34.  There are other issues remaining with scenes.  This is a very targeted update that does not address all of them.  One item to check:  I don't understand the extraction of `group` from `data` that some devices are doing.  I templated these changes from the working `handle_set()` method.  I noticed that Switch and Dimmer also attempt to get `group` from `data` in their `handle_scene()` methods.  Is this really correct?  Should KeypadLinc also do this?  ...only if the `group` argument to the method is `None`?

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -6,6 +6,7 @@
 - Fixed a bug where the MQTT client might close in an area that triggered an
   exception instead of a reconnect ([Issue #32][I32]).
 
+- KeypadLinc was using the wrong handler for scene commands ([Issue #34][I34]).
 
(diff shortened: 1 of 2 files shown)


TD22057:
Looks good - thanks.  The HISTORY file isn't quite right (needs a footnote link on the bottom) but I'll update that after the merge.

Issue #17, opened 2017-12-23
Support 5+ year old devices


TD22057:
See Issue #14 for some details (from @mwdng).  Old devices don't respond to the refresh command properly.  It may be that they need a standard cmd, not an extended one.

krkeegan:
I have a bunch of i1 devices so I am more than happy to try and tackle this.  This will have to go in stages.

My first issue is, in order for this to work, we need to know the "engine version" of the devices early on.  This is a trivial get_engine_version 0x0D standard command.  

My first question is, should we send this command and retrieve this information on the device config load, in from_config(), should we wait until we try and send a message that requires knowing the engine version, or should this be an affirmative command sent by the user?

TD22057:
Off the top of my head, I'd say something like:
- add this as a field to the device database (device.Db) and add it to the to/from json conversions
- add a method in device.Base to send this request and process the response by setting the value into the device db
- in the device refresh() command, if this field isn't set, send the command to get it

We know refresh has to be called at least once to get the link database so things work.  This would insure that the engine version is only requested one time per device.  After that, the devices can use it to switch behavior by looking at the field in the database.

krkeegan:
awesome that is exactly what I needed, thank you

krkeegan:
Alright, I made a pull request for this stage.

I do think that refresh is a good place for this. But the one thing to note, is that at least in the instructions, when adding a new device we will need to add a step for refresh before linking.

TD22057:
Then this approach isn't going to work.  You can't refresh before linking.  The device won't accept commands from the modem until it's linked.

krkeegan:
Hmm, ok the alternative is to force a get_engine request within the linking command if the engine version is not known.

krkeegan:
I have added support for writing link entries using the standard messages required by i1 devices.   It is in branch https://github.com/krkeegan/insteon-mqtt/tree/WriteDBForI1, I am going to wait on submitting a PR as many of the PRs and branches are overlapping.

I think this concludes the necessary improvements to close this issue when merged.

TD22057:
I believe this is now fixed w/ the merge of your PR's.  Let me know if that's not the case.

TD22057:
For completeness, that was PR #72, #41, and #40.

PR #40, opened 2018-03-09
Add Support for Engine Version

krkeegan:
Adds a get_engine command that can be used to get the engine
version which is saved in the device.db.

~~When refresh is called, if the engine version is not known,
it will be requested.~~

One step in process to fix https://github.com/TD22057/insteon-mqtt/issues/17

### DIFF
--- a/insteon_mqtt/cmd_line/device.py
+++ b/insteon_mqtt/cmd_line/device.py
@@ -71,6 +71,7 @@ def get_engine(args, config):
     reply = util.send(config, topic, payload, False)
     return reply["status"]
 
+
 #===========================================================================
 def print_db(args, config):
     topic = "%s/%s" % (args.topic, args.address)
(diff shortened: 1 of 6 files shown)


krkeegan:
I removed the changes to the device refresh process, as discussed this won't work.  The remaining additions for supporting a get_engine request are good.  At least in the interim, I can manually call get_engine on devices I know to be i1

TD22057:
Thanks for the contributions.  I'll make some time to integrate these but it's going to be a few weeks until I can get to them since I'm swamped at work with an upcoming delivery (and home with remodeling project) until the end of the month.  Just wanted to let you know that I'm not ignoring these...

krkeegan:
No problem, I will try and compartmentalize everything but some improvements overlap with each other.

TD22057:
Thanks.  I pushed a couple of minor tweaks and added a note in the release notes in HISTORY.md.

PR #41, opened 2018-03-10
Add Support for Refresh for I1 Devices

krkeegan:
This requires the engine support in PR #40 so this branch is branched off of that one.

Refresh as an i1 device will only occur if the device engine is known to be i1.  Otherwise default is to use standard i2/i2cs type commands.

Scanning a db using standard length messages is a bit messy, so this requires adding a DeviceScanManagerI1 to manage and send all of the necessary requests.

~~Once a full db record is received it is passed to DeviceEntryI1 which inherits almost all of its functions from DeviceEntry, however, the format of the record is slightly different.  This could also be accomplished by adding a staticmethod from_i1_bytes to DeviceEntry whatever the preference is.~~


krkeegan:
I thought about it more having an extra DeviceEntryI1 class was unnecessary, I removed it and condensed things.

krkeegan:
I resolved the conflict.  This is safe to merge.

krkeegan:
Discovered a change I had to make to use the device send function.

Now this is really ready to merge

PR #44, opened 2018-03-11
Catch and Ignore Duplicate Incoming Messages

### LINKED ISSUE #42: Add Message De-Duplication
I see that this is mentioned as a future enhancement in the docs already.  This is a bit more of an issue for i1 devices, particularly with database management.  This is because i1 devices send numerous standard length messages to get a complete link record.  To compound the problem, the messages sent from the device all look identical with the exception of the single data byte requested.  There is no state information.

I have previously written de-duplication processes that work decently well.  They look at the following factors:
- Cache recently received messages 
  - Messages are saved in the cache until the amount of time for any remaining hops_left has passed.  (87 milliseconds for standard length message hops and 187 milliseconds for extended length message hops seems to empirically be right, even though it doesn't match the insteon spec)
- Compare all incoming messages and ignore with warning duplicate messages
  - Messages are declared to be duplicates if everything about the message is the same with the exception of the max_hops and hops_left flags

krkeegan:
Keeps a list of messages that have not expired.  A non-expired message is one in which the time for the remaining hops to occur has not yet passed.

If a subsequent incoming message matches one of these, it is ignored with an info message to the log.

Fixes #42

### DIFF
--- a/insteon_mqtt/Protocol.py
+++ b/insteon_mqtt/Protocol.py
@@ -3,6 +3,7 @@
 # Insteon Protocol class.  Parses PLM data and writes messages.
 #
 #===========================================================================
+import time
 from . import log
 from . import message as Msg
 #from . import util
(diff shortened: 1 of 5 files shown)


TD22057:
FYI - I'm going to make a few tweaks to this after the merge.  I think using __eq__ on the Flags object and Inp/Exp objects is a little better than a duplicate method.  And adding __eq__ to Flags checks the extended flag which the current code doesn't (since extended inherits from standard, the isinstance isn't sufficient).

TD22057:
Using __eq__ I think worked out well.  The Protocol code becomes much simpler because it can just say "is msg in history_list" and it works.  Thanks for the tests - it makes making updates like this much nicer (I really need to get back to adding more unit tests for the other features...)

krkeegan:
Yeah that looks cleaner

PR #57, opened 2018-03-22
More leak sensor topic fix

djfjeff:
Now target the dev branch to make sure the fix will also be in upcoming release.

I also saw that the /leak topic was used in insteon_mqtt/mqtt/Leak.py. To be honest, I don't know why the /leak topic was not working since it seems to be registered correctly.

In any case, everything should be on the /wet topic for now (config, Leak.py and documentation)


djfjeff:
I will submit a more complete PR with support for heartbeat soon, I will close this one for now.

PR #60, opened 2018-03-26
Add support for Leak Sensor heartbeat

djfjeff:
The new code has been tested over the weekend, dry and wet event still work OK and heartbeat is reported every 24 hours.


TD22057:
Thanks!  I also added a line to the release notes in HISTORY.md.

PR #108, opened 2019-02-07
Call Base handle_broadcast; Revert change to Thermo Mode enum

### LINKED ISSUE #107: Thermostat cleanup
@krkeegan: I'm working on cleaning up some of the and making sure everything has correctly formatted doc strings and is using consistent naming conventions and methods.  I started in the network/, handlers/ and mqtt/ directories and will continue on to the devices/ soon.

I ran into a few things I'd like to change in Thermostat so I thought this would be a good place to communicate them.  
- mqtt/Thermostat: 
  - use default lower case for template args ('on' instead of 'ON') to match other devices.  I updated the sample config.yaml to add .upper() so that output is the same.
  - simplified template data specifications (one affect is that it now prefers temp_c over temp_f but I doubt that's a big deal)
- device/Thermostat: 
  - changed all enum values to be upper case to match other enums
- handler/ThermostatCmd: 
  - removed duplicate mode enum - changed code to use Thermostat.Mode

When you have a little time, please take a look at the cleanup branch and let me know if you see any issues with this.  You can ignore most of the diffs (lots of comment updates) - just check the thermostat files (found out that if you Alt-click the collapse file button, it collapse all of them which makes it a lot easier).  Be sure to set the diff to be against the dev branch.  

There is a TODO comment in device/Thermostat.py - I don't think the return statement should be there in the handle_broadcast() method.  At least none of the other devices skips the call to the base class.  That call is where linked devices are notified of the broadcast message so I think that return should be removed.

krkeegan:
Should always call handle_broadcast() in base to make sure linked
devices are updated.

The Mode enum in ThermostatCmd is necessary, the mapping of the
values are not the same for the various Insteon functions.

Fixes #107

### DIFF
--- a/insteon_mqtt/device/Thermostat.py
+++ b/insteon_mqtt/device/Thermostat.py
@@ -446,7 +446,6 @@ class Thermostat(Base):
                     status = Thermostat.Status.COOLING
 
                 self.signal_status_change.emit(self, status)
-                return  # TODO: should these be here?
 
         # As long as there is no errors (which return above), call
         # handle_broadcast for any device that we're the controller of.
--- a/insteon_mqtt/handler/ThermostatCmd.py
+++ b/insteon_mqtt/handler/ThermostatCmd.py
@@ -24,6 +24,16 @@ class ThermostatCmd(Base):
     NOTE: This handler is designed to always be active - it never returns
     FINISHED.
     """
+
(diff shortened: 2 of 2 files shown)


TD22057:
I've manually merged the changes into the cleanup branch.

PR #115, opened 2019-02-16
Keypadlinc disable toggle

jrevans:
Added capability to change keypadlinc button behavior so that pushing a button does not toggle the state but will emit a user defined signal of either 'on' or 'off'.


jrevans:

PR #117, opened 2019-02-16
Keypadlinc pair buttons

jrevans:
Added capability to pair keypadlinc buttons together (or unpair them).


jrevans:

PR #118, opened 2019-02-16
Keypadlinc Detach load

jrevans:
Added capability to detach the load on a Keypad link from button 1 and assign it to logical group 9.


jrevans:

PR #121, opened 2019-02-18
Device Category and Product enum

mooshee:
I creating enums so that product names and models could be printed alongside Insteon addresses.  This is just the start as it only prints it out in response to the `get_model` command.  I'd like to get feedback on where and when to print this so the logging is more helpful.

I scraped the data using a PDF to CSV parser from this document http://cache.insteon.com/pdf/INSTEON_DevCats_and_Product_Keys_20081008.pdf

I only have a little python experience so please scrutinize the code and style.

### DIFF
--- a/insteon_mqtt/device/Base.py
+++ b/insteon_mqtt/device/Base.py
@@ -6,6 +6,7 @@
 import json
 import os.path
 from .MsgHistory import MsgHistory
+from .Product import Product
 from ..Address import Address
 from ..CommandSeq import CommandSeq
 from .. import db
@@ -806,8 +807,8 @@ class Base:
             self.db.set_sub_cat(msg.to_addr.ids[1])
             self.db.set_firmware(msg.to_addr.ids[2])
             LOG.ui("Device %s received model information, dev_cat: %#x, " +
-                   "sub_cat: %#x, firmware: %#x ", self.addr, self.db.dev_cat,
-                   self.db.sub_cat, self.db.firmware)
+                   "sub_cat: %#x, firmware: %#x product: %s", self.addr, self.db.dev_cat,
+                   self.db.sub_cat, self.db.firmware, Product((self.db.dev_cat, self.db.sub_cat)))
             on_done(True, "Operation complete", None)
         else:
             LOG.debug("Device %s get_model response with wrong cmd %s",
--- a/insteon_mqtt/device/Base.py
+++ b/insteon_mqtt/device/Base.py
@@ -807,8 +807,9 @@ class Base:
             self.db.set_sub_cat(msg.to_addr.ids[1])
             self.db.set_firmware(msg.to_addr.ids[2])
             LOG.ui("Device %s received model information, dev_cat: %#x, " +
-                   "sub_cat: %#x, firmware: %#x product: %s", self.addr, self.db.dev_cat,
-                   self.db.sub_cat, self.db.firmware, Product((self.db.dev_cat, self.db.sub_cat)))
+                   "sub_cat: %#x, firmware: %#x - %s", self.addr, 
+                   self.db.dev_cat, self.db.sub_cat, self.db.firmware, 
+                   Product((self.db.dev_cat, self.db.sub_cat)))
             on_done(True, "Operation complete", None)
         else:
             LOG.debug("Device %s get_model response with wrong cmd %s",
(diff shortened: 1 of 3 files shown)


TD22057:
Thanks.  In the future, it's generally better to open an issue ticket first to discuss what the problem is before starting.  I'm not sure I understand what problem this solves - the devices already know what they are (Switch, Outlet, etc) and generally that is shown in most of the logs.  Maybe it would help if I understood what you feel was deficient in the current logs.

If we wanted to support an automatic device discovery system (download modem db and automatically create devices based on dev_cat) then I could see the need for this information - but I'd assume that would be useful only one time when someone starts using an existing Insteon setup.  When I was originally designing this system, I found several references that stated that the table of data is out of date (Insteon doesn't do a very good job with publishing info) and it was more useful to have meaningful short names ("deck", "kitchen") so the user would probably have to edit the config anyway.

mooshee:
Agreed on creating an issue ticket first.  I was mostly getting my feet wet, so I picked something easy to add.  I was hoping to create a Draft PR but didn't see the option.

1. Logging - I'm using Hassio, so debugging my setup is a bit cumbersome refreshing the logs view.  My biggest pain point is verifying that links are being setup correctly.  Typically, I run 'print_db' which spits out the Insteon IDs and then lookup my devices.  I believe `get_devices` already shows the labels / short names, but it would be nice if one or both of these also spit out the Insteon model.  In one instance, It took me an hour to realize I put a device in the wrong device type list in the config.yaml.  So perhaps we could verify the device types from the config.yaml against the Product enum (create groupings for switches, keypads, etc).

2. New setups - Yes, it would be very helpful for me as I configure my existing Insteon setup.  Others have expressed wanting to make the jump, but the onerous of tracking down and mapping all of your devices can be prohibitive (although most of it is unavoidable).  I was able to make sense of my current setup for porting to Insteon-MQTT by long-pressing each device in HomeKit and writing down device ID and model.  So this would also be useful for exporting setups as well.  Improving these experience is great for increased adoption.

3. Out of date table - This is meant to be a starting point.  We won't need the table going forward.  As new products are released (rarely) or found, we can easily append to the list.  The Insteon Product Keys (IPK) aren't particularly useful, I just added it because it was there.

4. Short names - Agreed.  Again, I really only see this useful for debugging / logging verbosity at the moment.

krkeegan:
I don't see this as necessary, but I can see it providing some help in a few limited cases.  I highly doubt the list there is complete.  As long as we are using this for user hints and not as part of an automatic system, completeness isn't essential.

mooshee:
I found [a better list](https://insteon.atlassian.net/wiki/spaces/IKB/pages/13533225/Insteon+Device+Categories+and+Sub-Categories) of Insteon products and dropped the product keys.  This information will be really important for supporting auto-discovery.  In the meantime, it's just a nice to have.

Here are some examples
```
$ insteon-mqtt config.yaml get-model 27.fc.43
Commanding fan_linc device 27.fc.43 (fan2) cmd=get_model
Device 27.fc.43 received model information, dev_cat: 0x1, sub_cat: 0x2e, firmware: 0x41 - DIMMABLE_LIGHTING [2475F] FanLinc
Operation complete

$ insteon-mqtt venv/config.yaml get-model 45.7b.12
Commanding keypad_linc device 45.7b.12 (kp1) cmd=get_model
Device 45.7b.12 received model information, dev_cat: 0x1, sub_cat: 0x42, firmware: 0x45 - DIMMABLE_LIGHTING [2334-232] Keypad Dimmer Dual-Band, 6 Button
Operation complete
```

TD22057:
I've merged this in using a different implementation (see 24a742a1e40a756939e07814e8ec631d6d17b062).  I don't like using an enumeration with lots of values - it requires making up enum names (like INSTEON_00_04) that never get used anywhere.  I find that anytime I'm going things like that, it means the design probably isn't the right one for that problem.  I added catalog.py which has a Category enumeration and a nested dict of [dev_cat][sub_cat].  When queried, these get turned into a catalog.Entry object which has all 4 pieces of info that you coded up in the enum (dev,sub,model,desc).   I then updated Device.Db to hold an Entry object instead of the dev_cat and sub_cat fields and changed the device.Base to use that to create the model info message instead of formatting it there and added unit tests for this new code.  This yields something like this:
```
Commanding keypad_linc device 46.7b.bc (porch) cmd=get_model
Device 46.7b.bc received model information: DIMMABLE_LIGHTING (0x01): '2334-232' (0x42) 'Keypad Dimmer Dual-Band, 6 Button' firmware: 0x45
Operation complete
```

Thanks for putting the catalog together and pushing this.

PR #143, opened 2019-05-15
Keypadlinc toggle enable/disable feature

jrevans:


TD22057:
This has been merged into the dev branch.  I changed to code to have a separate inputs in set_flags for "signal_bits" and "nontoggle_bits" which are the two low level commands.  The proposed api  in the patch doesn't work since it used None which is python only (hard to do in MQTT and command line).

Note: this really isn't finished.  The mask  state needs to be stored in the db and read at startup and checked during a refresh.  I'll work on that on the dev branch later.  I also want to add a higher level command to create radio button groups which will run all the right commands for someone.

PR #144, opened 2019-05-15
Keypadlinc radio button feature

jrevans:


TD22057:
This has been merged into the dev branch.  

Note: this really isn't finished.  The mask  state needs to be stored in the db and read at startup and checked during a refresh.  I'll work on that on the dev branch later.

PR #146, opened 2019-05-15
Keypadlinc load detach feature

jrevans:


TD22057:
Did you get this to actually work?  I thought we couldn't figure out how to command the load once it was detached.

TD22057:
This has been merged into the dev branch.   

Note: this really isn't finished.  The detach state needs to be stored in the db and read at startup and checked during a refresh.  I'll work on that on the dev branch later.

Also, there is no way to control the group 1 led in detached mode.  The only way I could find is to create a virtual modem scene linked to group 1 and toggle that.  That will have to be done automatically, saved in the db, and searched for on a cold startup to find the virtual scene number.

PR #147, opened 2019-05-15
Keypadlinc/Switch turn LED indicator on/off feature

jrevans:


TD22057:
This has been merged into the dev branch.  I changed the code so that the standard set_backlight() is used.  If the input is 0, it calls the special "turn off the light" command and records that info.  Then if a non-zero level is input, it sets that and if the light is off, it runs the "turn on the light" command.

Note: this really isn't finished.  The light on/off state needs to be stored in the db and read at startup and checked during a refresh.  I'll work on that on the dev branch later.

Issue #25, opened 2018-01-03
Create Insteon scenes in the config file.


bfranske:
I think about the last thing preventing me from fully switching over to insteon-mqtt is support for creating links and scenes between Insteon devices programmatically (e.g. through config.yaml). This effectively allows the creation of 3-way and 4-way switches from Insteon devices. For example this code (below) in Mr. House links together 3 switches in the kitchen to act as a 3-way switch circuit and 4 switches in the living room to act as a 4-way switch circuit controlling a lamplinc as well as the devices attached to those switches. Finally, these are also setup as virtual PLM buttons which can control the scenes as well.

```#Virtual PLM Buttons
#PLM Scene Controller,  CtlNum, Scene Name,             Group
INSTEON_ICONTROLLER,    02,     kitchen_lts,            All_Lights
INSTEON_ICONTROLLER,    03,     living_room_lts,        All_Lights


#               name,                   device,                         cont,   resp,   on_lev, ramp_rate
SCENE_BUILD,    kitchen_lts,            kitchen_ceiling_light,          1,      1,      100%,   0.5s
SCENE_BUILD,    kitchen_lts,            kitchen_kpl_main,               1,      1,      100%,   0.5s
SCENE_BUILD,    kitchen_lts,            kitchen_remotesw,               1,      1,      100%,   0.5s
SCENE_BUILD,    living_room_lts,        living_room_lamp,               1,      1,      100%,   0s
SCENE_BUILD,    living_room_lts,        living_room_lamp_remotesw1,     1,      1,      100%,   0s
SCENE_BUILD,    living_room_lts,        living_room_lamp_remotesw2,     1,      1,      100%,   0s
SCENE_BUILD,    living_room_lts,        kitchen_kpl_b,                  1,      1,      100%,   0s
SCENE_BUILD,    living_room_lts,        living_room_table_lamp,         0,      1,      100%,   0s
```
In this the "name " column refers to the Insteon scene name (used internally by Mr House only), the "device" column refers to the device name in the Mr. House configuration (which then provides the hex address), the "cont" column specifies whether the device should be a controller for the scene, the "resp" whether the device should be a responder in the scene, etc.

TD22057:
[...]

PR #158, opened 2019-08-02
Convert to Ruamel Yaml Support; Add Yaml Save Function

krkeegan:
This adds a save function which allows modifications to the Yaml config file made by the code to be saved to disk.  This is necessary in order for my implementation of scene management #25 to work.

Specifically this pull request contains the following:
- switches from Yaml to Ruamel
- Converts the config.py file and functions to a Config class.  The config object is then passed around to the various other objects.
- Adds a save function to the Config class.  A backup file is created if the saved file appears to be significantly different from the original file (hopefully this ensures that no user data is lost if a bug creeps in.)
- Adds unit tests for the save function.

A NOTE ON A SPECIFIC LIMITATION:
This code will read the !include tags and import them just fine.  However on export, it will compress the entire config code into the single parent config file removing the !include tags.  There is no loss of data, only a loss of organizational structure.

Trying to build in the ability for Ruamel to write out the separate include files back into individual files is hard and a bit ugly.  A testing file that I was using can be found here: https://github.com/krkeegan/insteon-mqtt/blob/scene_sync_ruamel_include/test-ruamel-include.py I believe the feature works, but it is super ugly and I would be afraid that it could have some hidden bugs.


krkeegan:
I am still working on this slowly.

I was thinking about this more.  Maybe it would be better to move the scenes definitions into their own file.  Say scenes.json.  Then I can use Ruamel on that file to alter and save the file. This would mean that !include could be used again on config.json.  It would also mean that any risk of loss of data from saving would be compartmentalized to the scenes definitions only.

krkeegan:
Yeah, I moved the scene definition to it's own file and this makes things much cleaner.  You can see it here: [scene_sync](/krkeegan/insteon-mqtt/tree/scene_sync)

I still have work to do, but I feel like this is making progress.  I am closing this pull request as the wrong implementation.

PR #183, opened 2020-01-28
Handle Housekeeping Messages Sent to Modem During Scene Command

krkeegan:
Solves issues with:

1. Timeouts because of a lot of AllLink Cleanup messages coming in.

2. Better handle failed scene commands by changing the successful devices.

3. Properly warn users about failures with error messages.

Ironically, this includes fixing a bug that is caused by an error in the Insteon documentation.  The same bug I fixed almost exactly seven years ago in MisterHouse.  Seven years working on Insteon, ugh, I would never have expected that.

### DIFF
--- a/insteon_mqtt/Modem.py
+++ b/insteon_mqtt/Modem.py
@@ -621,7 +621,7 @@ class Modem:
 
         cmd1 = 0x11 if is_on else 0x13
         msg = Msg.OutModemScene(group, cmd1, 0x00)
-        msg_handler = handler.ModemScene(self, msg, on_done, num_retry=1)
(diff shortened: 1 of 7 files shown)


krkeegan:
I pushed a few more commits that improve ALLLink commands even more.

Notably I did a lot more research on the Modem Scene process and most importantly, is that an InpAllLinkStatus NAK is not the final attempt by the Modem.  Additionally, we should increase the wait_time even when we receive garbled messages.

Second, I also added a delay after a device sends a Broadcast message.  We can calculate how many messages the device will send based off how many responders are present.  We can then cause a delay to avoid colliding with those messages.

All of this has resulted in a much more stable and accurate network when sending long and complex ALLLink Commands.

PR #196, opened 2020-03-02
Delete Entry from Group List; Delete List if Empty

krkeegan:
A simple bug fix.

The modem delete db function didn't handle group deletion properly.

This fixes that.

### DIFF
--- a/insteon_mqtt/db/Modem.py
+++ b/insteon_mqtt/db/Modem.py
@@ -174,7 +174,11 @@ class Modem:
         """
         self.entries.remove(entry)
         if entry.is_controller:
-            del self.groups[entry.group]
+            responders = self.groups.setdefault(entry.group, [])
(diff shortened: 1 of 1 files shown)


TD22057:
I'm not sure what to do to trigger this but I'm going to assume you're correct about the bug.  I've changed your code from this:
```
         if entry.is_controller:
            responders = self.groups.setdefault(entry.group, [])
            if entry in responders:
                responders.remove(entry)
            if not responders and entry.group in self.groups:
                del self.groups[entry.group]
```

to be the following (I don't like adding the empty list if it's not needed)

```
        if entry.is_controller:
            responders = self.groups.get(entry.group)
            if responders:
                if entry in responders:
                    responders.remove(entry)

            elif entry.group in self.groups:
                del self.groups[entry.group]
```

PR #197, opened 2020-03-04
IOLinc Improvements

### LINKED ISSUE #64: IOLinc relay state not reported
I use the IOLinc with a water valve, so I use it in Latching mode with no use for the sensor.

Triggering the scene 20 on and off effectively open and close the relay, so my valve work.

However, the MQTT state topic always report OFF because it tracks the sensor state. Without having the relay state available on MQTT, I cannot add the IOLinc to HASS as it always stay in the OFF position (because it track the unused sensor state in my case).

krkeegan:
This request now accurately models the IOLinc.  It accurately tracks the state of the Sensor and Relay.
Fixes #64

Features
- Tracks sensor and relay states
- Enables proper set-flags functionality (I am not sure this was working before)
- Enable setting of momentary timeout
- Stores flag values in persistent storage
- Turns off relay after momentary time passes (if correct values are known from get-flags)
- Relay will not be reported as on when in Momentary_C and sensor value is not in correct state
- Relay will be reported as on when in relay_linked state
- Adds specific relay and sensor topics

BREAKING CHANGES
- scene topic and command line option has been removed.  The prior design didn't offer proper control over Data1 values.  Plus why use a one off solution when the modem scene and scenes feature fully covers this.  I added a bunch of documentation to make this transition as simple as possible.

Structural Changes
- Adds a TimedCall feature which enables scheduling of functional calls.  Used to change relay state back to off after momentary timeout has elapsed.


TD22057:
Thanks for this.  My IOLinc only responds 50% of the time and I don't feel like dropping another $75 to get one.  I don't know if you saw this but Protocol already has some timer support.  See https://github.com/TD22057/insteon-mqtt/blob/master/insteon_mqtt/Protocol.py#L175  - the after input allows setting the time a message will go out.  Would that remove the need for the TimedCall class?  I don't really like having the Modem store all these extra things (stack, timed_call) that are network artifacts and not really part of the modem.   What's the reason for having the timed_call in the modem and not having it be a new method in Protocol (or better - an extension of the current send() method like what's there now)?

The original IOLinc used bit flags to set everything at once.  Isn't that better than a call sequence pushing one byte at a time?  Or I am not understanding the set_flags code?

krkeegan:
The timed message feature in Protocol sadly would not solve the need.  TimedCall is used solely to make a function call to update the internal state of the relay, no messages are sent.

Both TimedCall and Stack are internal processes, neither of them sends anything on the network (so they can't be an extension of send()).  They can be moved anywhere, but they seemed about as home in the Modem class as they do in the Protocol class since they really have nothing to do with the Insteon network.  They are instead things that are used to manage the internal state of Insteon-Mqtt.

As to the set-flag function.  The old code didn't work on my IOLincs.  I would receive an ACK, but running get-flags again would show that nothing had changed.  The implementation in this pull was taken from the developer notes and is also the same way that the Insteon App on the Hub sets these values as well, I think it should work on all models.  

For changes in the momentary mode, the pull request has one more insteon message (set A,B,C) vs (get_flags then set_flags).  But for the relay_linked and sensor_reverse it is one less call since there is no need for a get_flags call.  If the old version was working for you, can you send me the get_model details from your iolinc and I can try and add in the logic on when each call type should be made depending on the version of the device.

krkeegan:
I don't really understand it, but the pytests that previously worked exploded.  Anyways, I have fixed them again, it seems like something changed in how mock.patch works, or I was doing it wrong from the beginning and it just worked.

Anyways this works again.

PR #213, opened 2020-05-21
Add ability to pass dimmer ramp rate via MQTT commands

tstabrawa:
This change adds support for a new "transition" MQTT parameter, which matches Home Assistant's standard service parameter for specifying a desired transition time.  Support for transitions is implemented using the "Light ON at Ramp Rate" and "Light OFF at Ramp Rate" commands.  Since not all dimmers support these commands, a new config.yaml parameter is added to indicate device support (on_off_ramp_supported: True).

**Tests performed:**

- [x] Confirmed no new flake8 errors (see attached test log - [ramp-rate-testing.txt](https://github.com/TD22057/insteon-mqtt/files/4660812/ramp-rate-testing.txt))
- [x] Confirmed no new pylint errors (see attached test log - [ramp-rate-testing.txt](https://github.com/TD22057/insteon-mqtt/files/4660812/ramp-rate-testing.txt))
- [x] Confirmed all old & new pytest tests are passing (see attached test log - [ramp-rate-testing.txt](https://github.com/TD22057/insteon-mqtt/files/4660812/ramp-rate-testing.txt))
- [x] Reviewed test coverage report for substantial new gaps.
- [x] Manual testing by passing hand-written MQTT via HA developer tools menu.  Confirmed expected commands sent to device depending on whether on_off_ramp_supported is True or not.
- [x] Manual testing by sending service event via HA developer tools menu.  Confirmed expected commands sent to device depending on whether on_off_ramp_supported is True or not.
- [x] Dog-food tested with HA version 0.109.x and 0.110.0 for ~3 days without problems.
- [x] Confirm that there are insteon devices that support this feature.
- [x] Confirm that this works in real world testing
**Caveats:**

- I realized shortly after finishing implementing this feature that exactly zero of my dimmers actually support the "Light ON at Ramp Rate" and "Light OFF at Ramp Rate" commands, so technically, I've never seen a dimmer actually respond to the commands being sent.  When sending these commands to my dimmers, I get any of the following behaviors:
  - Dimmer acknowledges the command, but doesn't do anything.  Insteon-MQTT updates the dimmer's state based on the command succeeding.
  - Dimmer sends a negative acknowledgement to the command.  Insteon-MQTT does not update the dimmer's state.
- Considering that dimmers that support these commands appear not to be particularly common, the on_off_ramp_supported parameter defaults to False.  This avoids unexpected behavior in the event that HomeAssistant sends a "transition" parameter to a dimmer that doesn't support the commands.
- As a result of not being able to test with real hardware, all of my testing has been unit tests based on the published documentation for these commands in the following references:
  - http://www.madreporite.com/insteon/commands.htm
  - http://www.madreporite.com/insteon/ramprate.htm
  - https://cache.insteon.com/pdf/INSTEON_Developers_Guide_20070816a.pdf
- **If anyone has a dimmer that supports these commands, please try out this branch and post a comment here with your results.**
- I'm not a native Python speaker, so please be gentle while code reviewing.  :-)

**Attachments:**

- [ramp-rate-testing.txt](https://github.com/TD22057/insteon-mqtt/files/4660812/ramp-rate-testing.txt)


insteon_mqtt/device/Base.py
```diff
@@ -60,7 +60,6 @@ def from_config(cls, values, protocol, modem, **kwargs):
         for config in values:
             # If it's a dict, it's got a nice name set.
             if isinstance(config, dict):
-                assert len(config) == 1
```
krkeegan, line 63:
```yaml
- 3a.29.84: 'lamp2'
  on_off_ramp_supported: True
```

While concise, I don't think this config design works.  There is no order to a dict in python and as such we can't be sure that the first key/val pair is the addr/name.  In which case, the line below this comment would fail.

So if we end up sticking with hardcoded configs (or if we think this functionality may be useful in the future) we need to fix this.

I used something like the following in scenes, the 3 examples of how a device may be defined in the yaml file.  My programmer side would prefer just to use the long version every time since there is no ambiguity, but I recognize that for user happiness we probably need all three and just have to deal with the extra code.
```yaml
- DE.AD.BE
- EF.01.02: Device2
- 03,04.05:
    name: Device3
    on_off_ramp_supported: True
```
if
1. the config is not a dict, it is just the addr.
2. if the config is a dict, it must be a dict of length 1
2a. In that case if the val is not a dict, it is assume to be the name
2b. if the val is a dict, it is expected to have a name key in the dict

config.yaml
```diff
@@ -71,8 +71,10 @@ insteon:
       - 37.2d.20: 'deck'
 
     # Dimming devices (outlets, wall switches, lamp modules, etc).
+    # Note: on_off_ramp_supported is assumed false if not specified.
     dimmer:
       - 3a.29.84: 'lamp2'
+        on_off_ramp_supported: True
```
krkeegan, line 77: Is there anyway to determine if a device supports this by knowing the device-category, sub-category, and firmware?

I think the preference would be to do as much for the user as possible.  This probably means that we would have to expose the Ramp topic for all dimmers, even those that don't support it, since the MQTT instance is initiated at the start.  But I think we could gracefully handle a Ramp command sent to a device that doesn't support it (log a message to the user, and the set on_off as otherwise desired).  It looks like you already have code to gracefully handle this situation, so we would just need to add a log message.

We would have to build a table of device capabilities, so we would likely need input from users on what devices have this feature.  With any luck there will be a clear line where dimming devices with sub-categories greater than XX will support this capability.

krkeegan (commented):
I am a little weary of adding support that we can't verify that it works.  I also would like to know that such devices exist.  The Insteon spec has defined features that have never been enabled on any device in the past, so the spec alone isn't a reliable source.

I added two comments.  I think there is a fundamental flaw in your config entry design.  I also wonder if we can programatically handle this for the user rather than require them to defined it in the config.

config.yaml
tstabrawa: > Is there anyway to determine if a device supports this by knowing the device-category, sub-category, and firmware?

I agree that it would be better to auto-detect support for these commands.  [This page](http://www.madreporite.com/insteon/ramprate.htm) suggests that "newer devices (e.g. LampLinc V4.1 and up)" support these commands, though it wasn't immediately clear to me how to go from this single statement to a reliable firmware version cutoff.

As far as I can tell from [this table](http://www.madreporite.com/insteon/Insteon_device_list.htm), it probably means the newer non-dual-band LampLincs support the commands.  All of my dimmers here are dual-band.  Maybe that means these commands were added for a while, then dropped in the dual-band LampLincs?

Do you happen to have any of these newer non-dual-band LampLinc modules?  Or have you come across any resources that would shed more light on which other devices should support these commands?

tstabrawa:
> I am a little weary of adding support that we can't verify that it works. I also would like to know that such devices exist. The Insteon spec has defined features that have never been enabled on any device in the past, so the spec alone isn't a reliable source.

I agree that we shouldn't merge this change without testing with actual devices.  I unfortunately, ended up in the position where I had implemented support for these commands + unit tests, only to try testing it in the real world and find that none of my devices actually support it.  For the devices that I have, some would just ignore the command (and not turn the light on) and others would respond with an error.  In an attempt to salvage what I had done, in case it could be of value to someone else, I added the changes in commit 23e6db54599deee752955a233e815cc128d405a3, so that these commands wouldn't be sent to devices unless they were known to work with them (by manually enabling the feature via config.yaml).

In posting this PR, I'm basically fishing for someone who has a device that supports the commands who can help justify the PR's inclusion in the code (by being willing to try it out and see if it works).  Assuming it works, I'd be happy to replace the config-based enabling with auto-detection as you suggested.  Otherwise, if there's just not sufficient interest in this feature, I'd probably not be too heartbroken if we just close the PR.

krkeegan:
Adding the following task to make this easier to understand at a glance:

- [ ] Confirm that there are insteon devices that support this feature.
- [ ] Confirm that this works in real world testing

tommycw1:
I was under the impression that only the key[ad dimmer supported this function.  the native openHAB Insteon binding supports this, but I've never actually tried to use it.

image


I have a few keypad dimmers and can try to use this feature if that helps.

tstabrawa:
> I have a few keypad dimmers and can try to use this feature if that helps.

Thanks for offering to try this out.  That would be helpful, especially if it turns out your device(s) support this command.

A couple things to watch out for:
- This PR is a bit old, so it's actually still based on a version near to 0.7.2 as a baseline.  You might need to reverse-out some config.yaml changes you've made since then if you don't want to break other devices.  Alternatively, while testing, you could edit config.yaml to omit all devices other than the keypad(s) that you're trying with this feature.
- As Kevin pointed out, there are some coding problems with how I implemented to ability to selectively enable this feature for some dimmers.  Probably the easiest way to avoid these problems is to test using commit 25a1764137f403328afeb1a608da1c842e2676ca instead, which has support for the Light ON/OFF at Ramp Rate commands, but tries to use it for all dimmers.

If this sounds too cumbersome to test in its current form, let me know and I can re-base it to the current dev branch.

tommycw1:
> I was under the impression that only the key[ad dimmer supported this function. the native openHAB Insteon binding supports this, but I've never actually tried to use it.
> 
> image
> 
> I have a few keypad dimmers and can try to use this feature if that helps.

So, today I tried the implementation of this in the native openHAB native insteon binding, its a little weird on how to tell OH what ramp to use, but I was able to verify this worked on my keypad dimmer.  

It is defined here on page 7 here:

http://cache.insteon.com/developer/2334-xxx2486xdev-042013-en.pdf

image

tommycw1:
> let me know and I can re-base it to the current dev branch.

Yeah, if you could rebase quickly, that would be cool.  I can test it out pretty quickly then.

tstabrawa:
> Yeah, if you could rebase quickly, that would be cool. I can test it out pretty quickly then.

Rebase is done.  I'd recommend testing with commit 60652ccc15cc7d740809de2e8ae3506a64922ab2 to avoid potential problems coding issues mentioned above.

tommycw1:
> Rebase is done. I'd recommend testing with commit [60652cc](https://github.com/TD22057/insteon-mqtt/commit/60652ccc15cc7d740809de2e8ae3506a64922ab2) to avoid potential problems coding issues mentioned above.

Got it and I am on the right commit.  Can you verify the topic and payload I should use to test for a keypad?

tommycw1:
> Can you verify the topic and payload I should use to test for a keypad?

Never mind - I figured it out.

I used:

`mosquitto_pub -h 192.168.1.200 -p 8883 -i "test-client" -d -t "insteon/28.a4.22/set/1" -m '{ "state" : "on", "brightness" : 255, "transition" : 100 }'`

Success - I transitioned from 0 to 255 over 100 seconds.  I verified a number of different transition times on two of my three keypads and they all worked.  One of these three isn't connected to a load, so I cannot (easily) directly verify and since it is the same model/firmware/engine, I don't want to pull it and put a DMM on it just to verify.

Here are the details on my keypads:

Keypad 1
```
{ "cmd" : "get_model" }
Device 28.a4.22 received model information: DIMMABLE_LIGHTING (0x01): '2334-232' (0x42) 'Keypad Dimmer Dual-Band, 6 Button' firmware: 0x43
{ "cmd" : "get_engine" }
Device 28.a4.22 engine version: i2c
```
Keypad 2
```
{ "cmd" : "get_model" }
Device 2b.ad.52 received model information: DIMMABLE_LIGHTING (0x01): '2334-232' (0x42) 'Keypad Dimmer Dual-Band, 6 Button' firmware: 0x43
{ "cmd" : "get_engine" }
Device 2b.ad.52 engine version: i2c
```

Keypad 3
```
{ "cmd" : "get_model" }
Device 2b.9f.fd received model information: DIMMABLE_LIGHTING (0x01): '2334-232' (0x42) 'Keypad Dimmer Dual-Band, 6 Button' firmware: 0x43
{ "cmd" : "get_engine" }
Device 2b.9f.fd engine version: i2c
```

krkeegan:
Well that is cool, kinda sad that it only seems to work on (some?) keypadlincs.  Your dimmer looks to be of recent vintage, so I doubt it works on any dimmers.

tstabrawa:
Thanks for testing it out.  I guess it never occurred to me to try it with a KeypadLinc / to expect that to work any better.  I do have one KeypadLinc, and it looks like it has a slightly newer firmware version than what you tested.  Unfortunately, it also doesn't have a load attached.  I could at least try sending the command and see if the KPL believes that it's turned on/off afterwards.  I'm not sure how practical it will be to put a multimeter on it without annoying my wife/kids.

`Device aa.bb.cc received model information: DIMMABLE_LIGHTING (0x01): '2334-232' (0x42) 'Keypad Dimmer Dual-Band, 6 Button' firmware: 0x45`

@krkeegan Do you have any older KPL's that you can try this on?  Maybe we remove the Dimmer code for this feature and only enable it for KPL's with known working sub-categories / firmware versions?

tommycw1:
As a point of reference, the native openHAB Insteon binding has had this feature included on keypads since 2015.  I've never heard anyone complain about it not working on any specific keypads but frankly I think it is a infrequently used feature.  I honestly don't know what I would use it for.  If I wanted to turn on my lights at anything other than the normal RR, I'd probably just create a scene.

In looking for some other info, I stumbled across this discussion from when it was added to OH back in the 1.x days:

https://github.com/openhab/openhab1-addons/pull/3621

They do not have this available on dimmers, only keypads.

krkeegan:
I have I think 5 KPLs, but I  replaced all of them in the last 3 years.

I saved an old dinosuar that is an i1 device.  It does not work on that one (not unexpected), its model is
```DIMMABLE_LIGHTING (0x01): '2486D' (0x09) 'KeypadLinc Dimmer' firmware: 0x2d```

**Comment by krkeegan** [comment]
Yeah the rest of mine are 2334 except for one non-Dimmer version.  My guess is that this works on all 2334 models.

**Comment by tstabrawa** [comment]
I pushed some more commits which revert the configuration mechanism that @krkeegan found problems with and replace it with auto-detection based on the device model.  At this time, we will only pass the "Light ON/OFF at Ramp Rate" command to KeypadLinc model 2334-222 and 2334-232 devices.

@tommycw1 If you could spare the time, I'd appreciate if you would please try out commit 9c0afc46e640c809e54ca17ab476075dcb1c3704 with your KPL's.

File: insteon_mqtt/device/Base.py (hunk 1 above)
Comment (tstabrawa) [reply insteon_mqtt/device/Base.py]: Resolving, as these config-related changes have been reverted in this PR.

File: config.yaml (hunk 2 above)
Comment (tstabrawa) [reply config.yaml]: Resolving as support for auto-detection based on device category/sub-category has been added as of commit 9c0afc46e640c809e54ca17ab476075dcb1c3704.

**Comment by tommycw1** [comment]
> @tommycw1 If you could spare the time, I'd appreciate if you would please try out commit [9c0afc4](https://github.com/TD22057/insteon-mqtt/commit/9c0afc46e640c809e54ca17ab476075dcb1c3704) with your KPL's.

Not sure what is going on but this isn't working.  The keypad responds on/off but only does so at the default RR.  Seems that the command isn't passing correctly like the detection of model code isn't correctly seeing the model of the keypad.  Is there something I need to do so insteon-mqtt knows what model the KPL is?

```
2021-01-06 18:49:01.501 INFO KeypadLinc: MQTT received button press 28.a4.22 (master bedroom keypad) = btn 1 at 100 normal 
2021-01-06 18:49:12.080 INFO KeypadLinc: KeypadLinc message insteon/28.a4.22/set/1 b'{ "state" : "on", "brightness" : 255, "transition" : 100  }'
2021-01-06 18:49:12.081 INFO KeypadLinc: KeypadLinc message insteon/28.a4.22/set/1 b'{ "state" : "on", "brightness" : 255, "transition" : 100  }'
2021-01-06 18:49:12.082 INFO KeypadLinc: KeypadLinc input command: {'cmd': 'on', 'level': 255, 'transition': 100}
2021-01-06 18:49:12.083 INFO KeypadLinc: KeypadLinc 28.a4.22 cmd: on grp 0 255
2021-01-06 18:49:12.084 INFO Protocol: Write message to modem: Std: 28.a4.22, Type.DIRECT mh:1 hl:1, 11 ff 
2021-01-06 18:49:12.093 INFO Protocol: Read 0x62: Std: 28.a4.22, Type.DIRECT mh:1 hl:1, 11 ff ack: True
2021-01-06 18:49:12.344 INFO Protocol: Read 0x50: Std: 28.a4.22->40.d1.6e Type.DIRECT_ACK mh:1 hl:0 cmd: 11 ff
2021-01-06 18:49:12.345 INFO KeypadLinc: Setting device 28.a4.22 (master bedroom keypad) grp=1 on=255 normal
2021-01-06 18:49:12.346 INFO KeypadLinc: MQTT received button press 28.a4.22 (master bedroom keypad) = btn 1 at 255 normal 
```

krkeegan:
run `get-model 28.a4.22` and then try again.

tommycw1:
> run `get-model 28.a4.22` and then try again.

Done and confirmed this now works on my KPL.  I might suggest adding some logging to clue people into why the light turns on w/o ramp if the device is unsupported rather than just ignoring the transition command. (note, I didn't have debig logging enabled only level 20, so maybe that is already there...)

tstabrawa:
Thanks again for testing.  I agree that additional logging is needed.  I'll see what I can to do add logging for when a ramp command is issued but not supported by the device or when the model isn't known.

krkeegan:
Sorry, I made a small conflict.  It seems an easy fix, not even sure why git couldn't catch that on it's own.  I can fix it unless @tstabrawa are you still planning on adding logging about the ramp rate not supported?

tstabrawa:
Yeah, I have the logging implemented, I just didn't push the commit yet, as I didn't have a chance to test it (it got too late last night).  I'll probably test/push it tonight.  It looks like only config.yaml is conflicting, so that shouldn't be too bad to fix.

tstabrawa:
Some example log messages in recent updates:

> 2021-01-08 19:13:11.702 INFO Dimmer: Light ON/OFF at Ramp Rate not supported with dimmers - ignoring ramp rate.
...
2021-01-08 19:50:48.780 INFO KeypadLinc: KeypadLinc model not in DB - ignoring ramp rate.  Use 'get_model aa.bb.cc' to retrieve.

(The latter would actually include the real device address.  I just put in a fake one for the example above.)

krkeegan:
Sorry about this, I am happy to make this change if you want.

The code has been a little lax about proper logging levels, so this isn't your fault.  But these messages should I think be `error` level.  I tried to create some guidance about [logging levels](https://github.com/TD22057/insteon-mqtt/blob/dev/docs/CONTRIBUTING.md#logging).  Please suggest changes if you think I have not been clear or if I am wrong with any of my classifications.

The biggest thing here, is that INFO messages only show up in the LOG.  The user interfaces (Cmd_Line, WebCLI, MQTT) only report error, warning, and UI messages.  In this case, since the user's command cannot be followed I think `error` is correct, but it could be `warning` too, I am not to picky.

Again, let me know, and I will make the changes after a merge if you like.

Thanks!

tstabrawa:
Updated log messages (note, I also changed the text of the KeypadLinc one - the redundancy was bothering me)

> 2021-01-09 14:43:34.478 ERROR Dimmer: Light ON/OFF at Ramp Rate not supported with dimmers - ignoring ramp rate.
...
2021-01-09 14:45:28.265 ERROR KeypadLinc: Model info not in DB - ignoring ramp rate.  Use 'get_model aa.bb.bb' to retrieve.

krkeegan:
awesome thank you1

PR #218, opened 2020-07-06
Add support for EZIO4O

embak:
This add support for Smartenit EZIO4O 4 relay output module. This PR implements the usual state, set and scene MQTT features for the 4 outputs. Usage is similar to the KeypadLinc buttons. 
Control of the device register flag is also implemented. 
Scene support is done through modem scene (ALL-Link-Broadcast messages). 
Like with KeypadLinc buttons, edit scenes.yaml to associate each EZIO4O output with a different modem group and other Insteon devices.


TD22057:
Thanks but I can't accept this as is.  It looks like your editor made changes to files that aren't related to the feature you're adding.  Please look at the diff and clean it up to only contain the code/doc/test changes required for this feature.  Sorry.

embak:
I will make the corrections. Thanks.

PR #219, opened 2020-08-01
Add support for EZIO4O

embak:
This add support for Smartenit EZIO4O 4 relay output module. This PR implements the usual state, set and scene MQTT features for the 4 outputs. Usage is similar to the KeypadLinc buttons.
Control of the device register flag is also implemented.
Scene support is done through modem scene (ALL-Link-Broadcast messages).
Like with KeypadLinc buttons, edit scenes.yaml to associate each EZIO4O output with a different modem group and other Insteon devices.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -124,6 +124,10 @@ insteon:
     thermostat:
     #  - aa.bb.cc: 'downstairs'
 
+    # EZIO4O 4 output relay modules
+    #ezio4o:
+    #  - aa.bb.cc: 'relays'
+
 #==========================================================================
 #
 # MQTT configuration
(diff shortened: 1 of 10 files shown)


embak:
This PR replace previous one ( #218 )

krkeegan (approved):
I did a cursory review.  I don't own one of these devices, so my review is limited to looking for things that may break other devices.  

This seems very directed towards just the EZIO4O device, so if you say it works, it is fine with me.

embak:
I took care that this PR won't affect other devices.
I use this EZIO4O device to control my pool lights. It worked without any problem for me this summer.

PR #220, opened 2020-08-13
Add configuration key for MQTT client ID

kpfleming:
The `id` key in the `mqtt` section of the configuration can be used to
supply a unique ID for each service instance.

When multiple insteon-mqtt services are connected to the same MQTT broker,
they could be configured to use separate topics, but not separate client IDs.
This resulted in the broker breaking the connections as the services connected
and reconnected.


TD22057:
Thanks.  I'm going to add a commented out id input in the sample config.yaml file.  That's the only doc's for the yaml entries right now - otherwise no one will know that this option exists.  I'll also add this into the release notes.

kpfleming:
You're very welcome! This project is fantastic, I've got two instances running on my network now (house and garage, using 2413U PLMs) and it's fast and easy to configure.

PR #227, opened 2020-11-17
Detect disconnections during poll() calls

### LINKED ISSUE #222: MQTT connection loss results in lockup/100% CPU usage
insteon-mqtt: from dev branch as of today
paho-mqtt: 1.5.1

Problem occurs with Python 3.8.6 and 3.9.0.

keepalive is set to 30 seconds in insteon-mqtt, and 60 seconds in Mosquitto.

If I watch the MQTT traffic I can see pings every 30 seconds, which get responses, and all is good. If I restart Mosquitto, I can see the TCP FIN/FIN ACK/ACK sequence indicating that the TCP connection has been torn down.

However, nothing shows up in in the insteon-mqtt log (log level set to 10), and all MQTT communication stops. No Insteon device changes are sent over MQTT, in fact insteon-mqtt appears to be 'hung' at that point because the log doesn't even show incoming Insteon device messages. On the network, there is zero MQTT traffic.

I left it in this state for more than 3 minutes, and nothing changed. Restarting insteon-mqtt immediately reconnects to the broker and all is well. While in this condition insteon-mqtt is consuming an entire CPU core of my Raspberry Pi, even though no communications are occurring.

kpfleming:
The Paho-MQTT client may not invoke the 'on_disconnect'
callback when used with an external loop, so check for connection
loss during calls to loop_misc() and handle it appropriately.

Closes #222.

### DIFF
--- a/insteon_mqtt/network/Mqtt.py
+++ b/insteon_mqtt/network/Mqtt.py
@@ -179,8 +179,10 @@ class Mqtt(Link):
             passed in so that all clients receive the same "current" time
             instead of each calling time.time() and getting a different value.
         """
-        # This is required to handle keepalive messages.
-        self.client.loop_misc()
+        # This is required to handle keepalive messages and detect disconnections.
+        rc = self.client.loop_misc()
+        if rc == paho.MQTT_ERR_NO_CONN:
+            self._on_disconnect(self.client, None, rc)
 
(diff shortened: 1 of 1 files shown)


krkeegan (approved):
We should merge this.  I have never seen this issue, but the fix seems reasonable enough.

Issue #207, opened 2020-04-14
Support ramp rate (feature request)


balthisar:
While playing with #206, I realize that there's currently no support for setting the ramp rate for local control of the switch with set_flags. Dimmers, for example, default to 0.5 seconds.

It looks like this could be a simple addition. Looking at `Dimmy.py`, for example, `set_on_level` looks like a perfect prototype for support. Simply set D2 to 0x05, and the new `set_ramp_level` (or whatever) should just work. That's the Insteon side; I'm not sure how to add support via MQTT or the command line.

I'd be willing to dig into it, but I would only promise support for dimmers, as I have nothing else to test against. It would also take me at least a few weeks, because I'm currently in the midst of a transition from Indigo to Insteon-MQTT + HASS, and I'm not really setup for Python development (no time like the near present to add PyCharm to my repertoire).

Of course, if this isn't a feature you feel worth adding, please feel free to close.

krkeegan:
Support Merged with #235

PR #235, opened 2020-12-02
Support for the default ramp rate setting for Dimmer and KeypadLinc

jordanrounds:
This adds support for setting the ramp rate using the set_flags command. It supports a value between 0 and 31. I've tested locally on real hardware and setting the rate works without any issue. The changes I've made are basically a duplicate of the set_on_level method.

### DIFF
--- a/insteon_mqtt/device/Dimmer.py
+++ b/insteon_mqtt/device/Dimmer.py
@@ -523,7 +523,7 @@ class Dimmer(Base):
 
     #-----------------------------------------------------------------------
     def set_ramp_rate(self, rate, on_done=None):
-        """Set the device default on level.
+        """Set the device default ramp rate.
 
         This changes the dimmer default ramp rate of how quickly the it
         will turn on or off. This rate can be between .1 seconds and up
--- a/insteon_mqtt/device/Dimmer.py
+++ b/insteon_mqtt/device/Dimmer.py
@@ -638,7 +638,7 @@ class Dimmer(Base):
     def handle_ramp_rate(self, msg, on_done):
         """Callback for handling set_ramp_rate() responses.
 
-        This is called when we get a response to the set_on_level() command.
+        This is called when we get a response to the set_ramp_rate() command.
         We don't need to do anything - just call the on_done callback with
         the status.
 
--- a/insteon_mqtt/device/Dimmer.py
+++ b/insteon_mqtt/device/Dimmer.py
@@ -526,21 +526,26 @@ class Dimmer(Base):
         """Set the device default ramp rate.
 
         This changes the dimmer default ramp rate of how quickly the it
-        will turn on or off. This rate can be between .1 seconds and up
         to 9 minutes.
 
         Args:
-          rate (int): The default ramp rate in the range [0,31]
+          rate (float): Ramp rate in in the range [0.1, 540] seconds
           on_done: Finished callback.  This is called when the command has
                    completed.  Signature is: on_done(success, msg, data)
         """
         LOG.info("Dimmer %s setting ramp rate to %s", self.label, rate)
 
+        data_3 = 0x1c #the default ramp rate is .5
+        for ramp_key, ramp_value in self.ramp_pretty.items():
+            if rate >= ramp_value:
+                data_3 = ramp_key
+                break
+
         # Extended message data - see Insteon dev guide p156.
         data = bytes([
             0x01,   # D1 must be group 0x01
             0x05,   # D2 set ramp rate when button is pressed
-            rate,  # D3 rate
+            data_3,  # D3 rate
             ] + [0x00] * 11)
 
         msg = Msg.OutExtended.direct(self.addr, 0x2e, 0x00, data)
@@ -593,7 +598,7 @@ class Dimmer(Base):
             seq.add(self.set_on_level, on_level)
 
         if FLAG_RAMP_RATE in kwargs:
-            rate = util.input_byte(kwargs, FLAG_RAMP_RATE)
+            rate = util.input_float(kwargs, FLAG_RAMP_RATE)
             seq.add(self.set_ramp_rate, rate)
 
         seq.run()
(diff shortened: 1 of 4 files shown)


insteon_mqtt/device/Dimmer.py
```diff
@@ -521,6 +521,35 @@ def set_on_level(self, level, on_done=None):
         msg_handler = handler.StandardCmd(msg, self.handle_on_level, on_done)
         self.send(msg, msg_handler)
 
+    #-----------------------------------------------------------------------
+    def set_ramp_rate(self, rate, on_done=None):
+        """Set the device default on level.
```
tstabrawa, line 526:
```suggestion
        """Set the device default ramp rate.
```

insteon_mqtt/device/Dimmer.py
```diff
@@ -600,6 +634,24 @@ def handle_on_level(self, msg, on_done):
         else:
             on_done(False, "Button on level failed", None)
 
+    #-----------------------------------------------------------------------
+    def handle_ramp_rate(self, msg, on_done):
+        """Callback for handling set_ramp_rate() responses.
+
+        This is called when we get a response to the set_on_level() command.
```
tstabrawa, line 641:
```suggestion
        This is called when we get a response to the set_ramp_rate() command.
```

insteon_mqtt/device/KeypadLinc.py
```diff
@@ -870,6 +870,36 @@ def set_on_level(self, level, on_done=None):
 
         self.send(msg, msg_handler)
 
+    #-----------------------------------------------------------------------
+    def set_ramp_rate(self, rate, on_done=None):
+        """Set the device default on level.
```
tstabrawa, line 875:
```suggestion
        """Set the device default ramp rate.
```

tstabrawa:
Closes #207

insteon_mqtt/device/KeypadLinc.py
```diff
@@ -870,6 +870,36 @@ def set_on_level(self, level, on_done=None):
...
+        This changes the dimmer default ramp rate of how quickly the it
+        will turn on or off. This rate can be between .1 seconds and up
+        to 9 minutes.
+
+        Args:
+          rate (int): The default ramp rate in the range [0,31]
+          on_done: Finished callback.  This is called when the command has
+                   completed.  Signature is: on_done(success, msg, data)
+        """
+        LOG.info("Dimmer %s setting ramp rate to %s", self.label, rate)
```
krkeegan, line 886: I suspect nothing bad would happen.  But it would be nice to check if this keypadlinc is dimmer capable.  Otherwise I don't think the ramp_rate is necessary or would work.

Something like would do it.
```python
if !self.is_dimmer:
  LOG.error("Can't set ramp rate device %s is not a dimmer", self.label)
  return
```

krkeegan (commented):
I note that this will only work with i2 or i2cs devices.  There is a way to do this for i1 devices, but I am starting to question whether we want to continue adding features for such devices.  They are very old now and often cause a lot of issues.  It is hard to test the code as most developers do not have these devices.

I would be content to start a type of "deprecation" of i1 devices.  We have the basic functionality for supporting them, I am not suggesting removing that.  But I would be fine with not adding support for additional features unless an owner of i1 devices wants to step forward and write the code.

I had one comment about checking to see if the KPL is a dimmer.  If we address that, then I see no reason not to merge this.

insteon_mqtt/device/KeypadLinc.py
jordanrounds: Good call! I added the check for is_dimmer like you suggested. I forgot some keypads aren't dimmers.

krkeegan (approved):

insteon_mqtt/device/KeypadLinc.py
```diff
@@ -870,6 +870,41 @@ def set_on_level(self, level, on_done=None):
...
+    #-----------------------------------------------------------------------
+    def set_ramp_rate(self, rate, on_done=None):
+        """Set the device default ramp rate.
+
+        This changes the dimmer default ramp rate of how quickly the it
+        will turn on or off. This rate can be between .1 seconds and up
+        to 9 minutes.
+
+        Args:
+          rate (int): The default ramp rate in the range [0,31]
```
tstabrawa, line 882: Sorry for the late comment.  I noticed that this code is expecting a ramp rate value that would be handed directly to the device -- so the user would be expected to know what value to use based on how those values are interpreted by devices (such as by consulting [this table](http://www.madreporite.com/insteon/ramprate.htm).)

I just wanted to raise the question - would it be better if the ramp rate were specified in seconds, and translated using the self.ramp_pretty lookup table, similar to as done in link_data_from_pretty()?

This might be a question for @krkeegan or @TD22057 to answer, though.

krkeegan:
Ahh, yes good catch, that is a better way to handle things.  The code is already available.  Look inside Devices/Dimmer you should see a dict of ramp_pretty.  Something like the following code should handle the conversion for you:

```python
data_3 = 0x1f
for ramp_key, ramp_value in self.ramp_pretty.items():
    if rate >= ramp_value:
        data_3 = ramp_key
        break
```

Then set your D3 value using data_3.  That code will select the exact ramp rate if it exists or the next lowest valid value.

jordanrounds:
Made the changes you suggested. Needed to add to the util to support float. Not exactly sure I handled everything necessary or am following the patterns elsewhere in the code, but that seemed like where I was supposed to be doing it.

krkeegan:
Bah, this was merged into master.  That was my fault, I should have looked more carefully.  I am quickly resetting master before anyone else syncs to it.

I will rebase it off of dev and resubmit it for you Jordan.  In the future, please create your branch off of dev.

jordanrounds:
Sorry about that. This is my first time contributing anything. I'll create the branch off of dev next time.

krkeegan:
No problem, it happens a lot, I should have caught it.  All fixed by #253 now.

PR #237, opened 2020-12-03
Don't Retry DB Get But Increase Timeout; Retry on DB Modify

krkeegan:
This issue was discovered in part by #231 

For some reason we had 0 retries on DB_Modify, which was wrong.  This changes that to the default 3.

_-- Update see comment below, we can have retry of 3 for the initial get message --_
Also, it became clear why the DB_Get handler must have a retry of 0.  The handler only sends a single request and then waits for numerous extended messages from the device.  If the time between messages from the device exceeds the time_out, there is really nothing we can do.  There is no way to ask the device to resume from that point.  In fact, asking again while the device is still responding seems to just be ignored by the device.

So instead, this increases the time_out for this particular handler.  It should decrease some errors with little side effect.  If the handler receives the final entry in the DB the handler exits with a complete and will not block for the full timeout.

### DIFF
--- a/insteon_mqtt/handler/DeviceDbGet.py
+++ b/insteon_mqtt/handler/DeviceDbGet.py
@@ -22,7 +22,7 @@ class DeviceDbGet(Base):
     Each reply is passed to the callback function set in the constructor
     which is usually a method on the device to update it's database.
     """
-    def __init__(self, device_db, on_done, num_retry=0):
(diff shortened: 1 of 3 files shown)


krkeegan:
I had a realization that the retries and timeout can still apply to the initial get request which is acked by the device.  Once the modem receives this ack, this change removes the retries and doubles the timeout since there is no further message that the modem can send.

This small change would help where the device simply misses the first get request.

Issue #142, opened 2019-04-29
Temperatures reported by Tstat in F being interpreted as C


azteknetworking:
Thanks for a great project.  Works pretty much perfect except for climate support. Two things here:

The set point state shows both C and F correctly, but it seems that while my insteon thermostats (2441TH) report their temps in F, the insteon-mqtt app is interpreting those as C, and then performing the calculation from C back to F, making the F temp incorrect.

Output from MQTT logs:
`insteon/downstairs/ambient_temp {"temp_f" : 156.0, "temp_c" : 68.88888888888889}
insteon/downstairs/cool_sp_state {"temp_f" : 77.0, "temp_c" : 25.0}`

Second, the reports from the Tstat aren't quite the same as what they show on the local Tstat screen:

`Apr 29 08:53:46 piserver insteon-mqtt[7766]: 2019-04-29 08:53:46 INFO Thermostat: MQTT received temp change 1e.0c.c7 (downstairs) = 72.77777777777777 C
Apr 29 08:54:47 piserver insteon-mqtt[7766]: 2019-04-29 08:54:47 INFO Protocol: Read 0x50: Std: 1e.0c.c7->44.86.b7 Type.DIRECT cmd: 6e 9c
Apr 29 08:54:47 piserver insteon-mqtt[7766]: 2019-04-29 08:54:47 INFO Thermostat: MQTT received temp change 1e.0c.c7 (downstairs) = 68.88888888888889 C
Apr 29 08:54:48 piserver insteon-mqtt[7766]: 2019-04-29 08:54:48 INFO Protocol: Read 0x50: Std: 1e.0c.c7->44.86.b7 Type.DIRECT cmd: 6f 1a`

But the local tstat shows the correct temperature of 78 and 80 degrees for the two time periods mentioned (I used a heat gun on the tstat to make it change and broadcast temp changes).  insteon-mqtt seems to think the temps were about 10 degrees lower than that.

woferry:
I don't think the issue is C vs F, what I'm seeing is that temp_f is reporting ~2x of the proper value.  I.e. when my 2441ZTH's screen shows 74F "temp_f" reports 147.0 (so presumably it's really 73.5F).  temp_c reports 63.888888888888886, which is proper math for the F to C conversion and is close to 74F, but that's just a coincidence I believe, the real issue seems to be that temp_f needs a divide-by-2, the 2441ZTH is probably reporting the temperature in half-F units rather than actual degrees.

[...]

PR #241, opened 2020-12-11
Fix Ambient Temp Calculation in Thermo; Add More Documentation

krkeegan:
Fixes TD22057\#142

Ambient Temp push messages are in 2x format.

Add get_status to normal refresh commands.  Ensures that we have
the proper units (C or F) set.  Otherwise we may misinterpret
temp messages.

Add more documentation to try and make things easier to understand.


krkeegan:
This is weird, it looks like from the web interface that this would commit changes.  But looking at the files manually, I can see that all of these changes are already committed.  I am very confused.

PR #244, opened 2020-12-11
Check the Remote Battery Periodically and Report to Low Battery Topic

### LINKED ISSUE #49: Support for battery reporting on Mini-Remote 4
It seems the Mini-Remote 4 does not send the battery heartbeat like other battery powered Insteon product. A workaround is to query the battery state when the remote is awake (after getting a button press for example).

I found this on the subject : https://krkeegan.com/insteon-remotelinc-2-lacks-the-heartbeat-feature/

krkeegan:
This requires merging #240 first.

This is a new feature which will check the battery level for 2342-xxx remotes at most once every 4 days.  The device is generally asleep, so the message is sent just after receiving a message from the device when it is awake.  This means you will not get battery warnings from devices that you do not use.

If the battery voltage is low, a message will be emitted to the Low_Battery topic.

- [x] Linting passes
- [x] Unit tests were updated and pass
- [x] Documentation was updated
- [x] Command line functions were added
- [x] Works well in a test environment, attempting to drain the battery on a remote, but actually testing would take a while

Closes #49

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -24,6 +24,11 @@
      manually awaken via holding the set button.  This will cause all queued
      and future messages to be sent to the device for up to three minutes
 
+- Added support for querying the battery on a mini-remote.  The battery state
+  will be automatically queried when the device wakes up if is has been 4 days
+  since the last battery check and will emit messages on the low_battery
+  topic. ([PR 244][P244])
+
 - Added support for Smartenit EZIO4O 4 relay output module (thanks @embak)
   ([PR 219][P219])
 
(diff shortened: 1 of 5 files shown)


krkeegan:
I am trying to determine what the low battery voltage should be.  I made a simple script that keeps querying the database trying to run out the battery.  So far I estimate it has sent 10,000 messages and still appears to be about 3.55v.  Once I have a good number this should be good to do as well.

krkeegan:
3.4v seems like a reasonable low battery level.  The device works down to at least 3.36v but the voltage seems to be dropping faster now, so it seems like this is the tail end of the curve.

As an aside, these things really run for a long time on a single charge.  The docs say 9-12 months.  While testing this, I figure there have been 20,000+ messages sent from this device and I have kept it awake for many hours now in order to drain it down.  The self discharge rate of Li-Ion is very good, so I would guess that a infrequently used device could last years, even a frequently used one probably lasts much longer than a year.

PR #256, opened 2020-12-15
Don't treat different broadcast msgs from same device as duplicate

tstabrawa:
This fixes a problem that I noticed recently where when some broadcast messages aren't received by my PLM and the corresponding cleanup message is received, the cleanup message is being dropped by Insteon-MQTT.  Upon digging in to the code, I noticed that the Broadcast handler had been written to attempt to avoid double-handling broadcasts and their corresponding cleanup messages by dropping any cleanup messages that follow a broadcast from the same device.  In the case of my i2cs devices, this meant that even if a Success Report Broadcast message from a previous button press had been received, then a cleanup message for a subsequent button press would be treated as a duplicate of the Success Report Broadcast message.

To fix, I made the Broadcast handler's duplicate check also consider whether the broadcast messages have the same command and group number.

**Tests performed:**  (see attached test log - [cleanup-fix-testing.txt](https://github.com/TD22057/insteon-mqtt/files/5693795/cleanup-fix-testing.txt))

- [x] Confirmed that all new pytest tests fail before fix and pass after fix
- [x] Confirmed no new flake8 errors
- [x] Confirmed no new pylint errors
- [x] Confirmed all old & new pytest tests are passing
- [x] Reviewed test coverage report for substantial new gaps
- [x] Confirmed fixed code works in real life
  - Using an i2cs plug-in dimmer module, turned device off and on using physical buttons
  - Observed that Success Report Broadcast message was received for the off command, broadcast message for on command was not received, and cleanup message was processed by Insteon-MQTT.
- [x] Dog-food testing (~24 hours with Home Assistant 0.118.5, 24 hours with Home Assistant 2020.12.0, no problems observed)

**Attachments:**

[cleanup-fix-testing.txt](https://github.com/TD22057/insteon-mqtt/files/5693795/cleanup-fix-testing.txt)

### DIFF
--- a/insteon_mqtt/handler/Broadcast.py
+++ b/insteon_mqtt/handler/Broadcast.py
@@ -83,7 +83,7 @@ class Broadcast(Base):
             self._last_broadcast = None
             return Msg.CONTINUE
 
-        # Different message flags than we exepcted.
+        # Different message flags than we expected.
         return Msg.UNKNOWN
 
     #-----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


krkeegan:
Good catch, that makes sense.

krkeegan (approved):

PR #285, opened 2020-12-28
Improve on-level handling for dimmers & related devices

tstabrawa:
Improve on-level handling for dimmers & related devices

- Store default on-level in DB metadata when set-flags commmand succeeds.
- Use stored default on-level to determine new level when receiving "on" broadcast messages.
- Use stored default on-level to handle MQTT messages with unspecified brightness value.
- Note: Previous behavior for handling MQTT messages with unspecified brightness value will remain for existing users, unless config is updated as shown in example config.yaml file).

**Tests performed:**

- [x] Confirmed that all new pytest tests fail before fix and pass after fix
- [x] Confirmed no new flake8 errors
- [x] Confirmed no new pylint errors
- [x] Confirmed all old & new pytest tests are passing
  * Some unrelated tests failed on my system (both before and after these changes were applied).  Old Python version might be to blame.
- [x] Reviewed test coverage report for substantial new gaps (no new gaps found)
- [x] Confirmed changes work in real life
  - Using an i2cs plug-in dimmer module and a KeypadLinc dimmer module:
    - Pressed physical on button several times.  Confirmed that brightness as reported in Home Assistant follows actual brightness of the connected light.
    - Turned device on in Home Assistant.  Observed device went to default on-level instead of full-brightness.
    - Manually sent MQTT to device with unspecified level.  Confirmed device alternated between default on-level and full-brightness like with physical buttons.
- [x] Dog-food testing (~1 week with an older version of these commits on Home Assistant 2020.12.0, no problems observed).  Current code testing is ongoing (~2 days so far with HA version 2020.12.1).

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -330,7 +330,7 @@ mqtt:
     # reason is input, is will be passed through to the state_payload.  The
     # output of passing the payload through the template must match the
     # following:
-    #   { "cmd" : "on"/"off", "level" : LEVEL,
+    #   { "cmd" : "on"/"off", ["level" : LEVEL],
     #                         ["mode" : 'normal'/'fast'/'instant'],
     #                         ["fast" : 1/0], ["instant" : 1/0] }
     # where:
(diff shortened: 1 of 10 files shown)


krkeegan:
This is great!  This was definitely needed.

In the process you managed to uncover some bad design I created that has been bugging me.  The Meta key that is saved in the database file looks like this:
```yaml
meta:
  IOLinc:
    momentary: true,
    other_value: false
```
Note that the IOlinc values are stored in a subkey of IOLinc.  The same is true for Motion, Remote, and Thermostat, which are the only objects that used set_meta.  I honestly don't know why I made a subkey like this.  The original discussion is here, https://github.com/TD22057/insteon-mqtt/issues/89.  I think I may have taken Ted's original idea too literally.  It is obviously much easier to just pass keys and values back and forth to set_meta and not dictionaries.  I don't think we get anything special from the extra subkey.

Anyways, your code does it in the simpler manner, I think this is the correct answer.

So, I will try and come up with some code that migrates in the old meta storage design for the existing devices into a flat meta storage like this.

tstabrawa:
Thanks for the kind words.  This one has been on my wish-list for a while, so it's satisfying to see it finished.  I'm thoroughly enjoying running this branch in production.  :-)

One thing that this code doesn't do (mostly because I wasn't aware that it's possible until today) is to read the current default on-level from devices that already have one set.  It looks like this can be done with an extended-get/set command (0x2e).  In your experience, is the ability to read the default on-level with extended-get consistently implemented?  Or perhaps by trying to read it, we'd end up reading a garbage value from some devices and storing that in the DB?

krkeegan:
Funny you say that, I coded that up.  I just need to finish the tests for
it tomorrow.

krkeegan:
I sent a PR to your branch that queries the on_level and ramp_rate as part of the get_flags action.

PR #295, opened 2020-12-31
Bound Backlight from 0x00 to 0x7f

### LINKED ISSUE #294: Bounds for LED Brightness incorrect
I noticed that when setting LED backlight brightness, setting to values < 16dec or 0x11 results in 0x11 still being sent to the device and the brightness not changing.

Looking at the code, I see the following note:

```
# Bound to 0x11 <= level <= 0xff per page 157 of insteon dev guide.
# 0x00 is used to disable the backlight so allow that explicitly.
```

Looking at the dev guide I see the following on page 157

image

Note: this bounds the upper limit to 0x**7**F, not 0xFF

I can also state that from experience, setting this to a value > 0x7f does not result in a brighter LED.

I also see this in the dev notes for the keypad and a micro dimmer module

image

image

There is also a problem with the lower bound.  The lower bound in these documents is stated to be 0x11. However I can achieve incrementally lower LED brightnesses all the way down from 0x11 to 0x00, using Insteon-Terminal (which does not implement a boundry).  This is the case for both the 2477D Dimmer as well as the 2334D Keypad.  

Lastly, setting the LED to 0x00 does NOT actually disable the LED, it just sets it to a very low level.  You need to set operating flag 9 for on or 8 for off:

image

From Insteon-Terminal:

```
        def setLEDOff(self):
                """setLEDOff()
                This disables the backlight LED"""
                self.querier.setMsgHandler(DefaultMsgHandler("Set LED OFF"))
                return self.querier.queryext(0x20, 0x08, [0x00, 0x00, 0x00]);

        def setLEDOn(self):
                """setLEDOn()
                This enables the backlight LED"""
                self.querier.setMsgHandler(DefaultMsgHandler("Set LED ON"))
                return self.querier.queryext(0x20, 0x09, [0x00, 0x00, 0x00]);
```

tommycw1:
Fixes #294

### DIFF
--- a/insteon_mqtt/device/Dimmer.py
+++ b/insteon_mqtt/device/Dimmer.py
@@ -473,10 +473,10 @@ class Dimmer(Base):
         """
         LOG.info("Dimmer %s setting backlight to %s", self.label, level)
 
-        # Bound to 0x11 <= level <= 0xff per page 157 of insteon dev guide.
-        # 0x00 is used to disable the backlight so allow that explicitly.
+        # Bound to 0x11 <= level <= 0x7f per page 157 of insteon dev guide.
+        # However in practice backlight can be incremented from 0x00 to 0x7f
         if level:
-            level = max(0x11, min(level, 0xff))
+            level = min(level, 0x7f)
 
         # Extended message data - see Insteon dev guide p156.
         data = bytes([


krkeegan:
Thanks for the patch.  Can you make the PR off of the dev branch?  If not, I can cherry pick and make a new PR for you.

tommycw1:
Sure, I should have thought of that. I’ll prob do it tomorrow

tommycw1:
Closing this request and adding to dev branch.

PR #297, opened 2021-01-01
Bound Backlight from 0x00 to 0x7f

tommycw1:
Signed-off-by: Tom Weichmann <tommycw10@gmail.com>
Fixes issue reported in #294


krkeegan:
Thank you for contributing this!

I tested this out.  My dimmer dims down to 0x04 at 0x03 it has a cliff, but it just barely on.  I can't see any change above 0x7F.  The same is true of my switches.  On Keypadlincs I can see changes all the way down to 0x01.  I can barely see any change on my outlet devices.

I made a new PR #299 with your changes plus some fixes and cleanup.

PR #303, opened 2021-01-05
Check the Modem Address on Startup

### LINKED ISSUE #266: Query PLM for its Address Rather than set in Config
Use the Get_IM_Info 0x60 command.

krkeegan:
This will check the modem address on startup:
If no address is specified in the config file:
1. It will use the address obtained from the modem; OR
2. Fail and exit if no address is obtained.

If an address is specified in config:
1. If the returned address is the same, nothing
2. If the returned address is different, an error is reported,
   and things will continue using the address in the config.

Closes #266 

- [x] Works in testing
- [x] Cleanup code documentation
- [x] Change config.yaml
- [x] Unit Tests

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -37,7 +37,10 @@ jobs:
     - name: Test with pytest
       run: |
         # Skips creating coverage stats for covered items
-        pytest --cache-clear --cov-report term:skip-covered --no-cov-on-fail --cov=insteon_mqtt tests/ | tee pytest-coverage.txt
+        # Set pipefail so that tee doesn't override the exit code
+        set -o pipefail
+        pytest --cache-clear --cov-report term:skip-covered --cov=insteon_mqtt tests/ | tee pytest-coverage.txt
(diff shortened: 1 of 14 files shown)


krkeegan:
I also added the `get_model` functionality to the modem since we get that information with the address request anyways.  

It could be useful for debugging or maybe something in the future.

jonsmirl:
I don't know about this one...

   If the returned address is different, an error is reported,
   and things will continue using the address in the config.

I messed up my entire house network by having the sample modem address uncommented in the config and of course that didn't match my actual modem address.   I had to repair every device again using the correct modem address.  I would suggest exiting if they disagree with a message telling the user to fix it. 

What would be a use case for wanting to use an address that didn't match your actual modem?

krkeegan:
I am worried about messing up current installations that use the address configuration.  The sample config file has been changed so that the default address is not enabled anymore.

jonsmirl:
But why would you want it not to match the actual modem? If you write the links in with the wrong modem address like I did, nothing works.

Issue #199, opened 2020-03-11
Finish Implementation of Reason Attribute


elperepat:
I updated insteon-mqtt and just seen that I could get the reason of a change of state. Great! It works for swiches and dimmers, but I can't get it to work completely for keypadlincs:

In my config.yaml I have:

`btn_state_payload: '{ "state": "{{on_str.upper()}}", "reason" : "{{reason}}" }'`

and I get the following mqtt message:

`{ "state": "ON", "reason" : "" }`

I'm expecting "command" like for the switches and dimmers

That's when I send a mqtt command to set the output. When I press the button on the keypadlinc, I get the correct reason: 

`{ "state": "ON", "reason" : "device" }`

Anything I could do to help?

Thanks for the great work!

Patrick

krkeegan:
That is because "reason" hasn't been incorporated everywhere in the code yet.  The default response is an empty string.  The command reason probably hasn't been passed through properly.

It would help if someone went through and diligently looked to see that anyplace where reason could be passed that it is for all of the devices.

krkeegan:
OK, I am starting to look at this.  This reason feature is complicated and maybe not well implemented in spots.  I see why this is happening with the KPL.  I need to think about whether I want to try and implement the reason feature in a better manner or if I should just patch this error.  I will have an answer tomorrow.

krkeegan:
Alright, the reason feature is fully implemented in the code, it should work everywhere now.  Please create an issue if you find a place where it is not working as expected.

PR #308, opened 2021-01-07
Refactor Scene Command; Fix Scene Command; Add Trait Concept

### LINKED ISSUE #283: Scene Function Relies on Receiving a Broadcast Report
Ted wrote the scene() function which commands the device to act as though it was manually pressed.  I had never seen use of this function prior to this.  I don't currently use it, so I can't say how well it functions.

But while reviewing the code, I noticed that the completion of the scene function relies on receiving the All_Link_Broadcast Cleanup_Report from the device.  I think this could be unreliable.  The message is a broadcast, so the device does not confirm that we have received it.  I suspect there are times we don't see it.  

Since I don't use the functionality I can't say how frequently a failure occurs, or what happens when it does.  But I am putting this here as a note to consider altering this.

krkeegan:
I am particularly happy with this.  It fixes #283, #307, and a portion of #199.

It dramatically simplifies the scene command, no longer is a followup call to off needed when sending the off command.  It removes a bunch of nested callbacks and odd logic.  Importantly it cuts a significant amount of heft from the code.

Fix the scene command so that when turning off, the device also turns off without needing an additional command.  Also fixes the issue where the device would turn on briefly when turned off if it was off originally.  Also removes the dependence on receiving the link-cleanup report.

Create the concept of trait classes that contain traits that are reused across some but not all devices.

Condense all scene code into a single trait class.  Add trait to Switch, Dimmer, KPL and FanLinc(via inheritance through Dimmer)

Update Unit tests, except still need work on FanLinc.

- [ ] Update fanlinc unit tests


krkeegan:
FanLinc, EZIO4O, and Outlets currently offer the scene command.  Yet, none of these devices have physical controls.  I suppose one could use the set button on the outlet, but that seems weird.  __Does it make sense to have the scene command for these devices?__  I suspect it was simply added because developers were copying files when adding new devices.

krkeegan:
I also discovered that a level can be sent with the scene command, which is nice because it makes scene basically a full replacement for the set or level topics for those who want to use it.

The tests still fail because FanLinc is using the scene topic.  @tstabrawa i know you don't speak for everyone, but I know you have a fanlinc.  Do you see anyone using the scene topic on the fanlinc?  See [here](https://github.com/TD22057/insteon-mqtt/blob/master/docs/mqtt.md#scene-triggering) for an explanation of what it does.  It seems to me, that since the fanlinc is not itself a controller (it really has no buttons) no one will be setting it up as a controller of another device, making this feature basically useless.

I think the same is true of outlet.

EZIO4O also has the scene enabled, but I think this was just a mistake in copying code and it seems pretty clear from the code notes that it doesn't belong.

tstabrawa:
I mostly use the FanLinc by controlling it from a KeypadLinc (both the fan and light loads are connected to the FanLinc, and the KeypadLinc has no connected load).  The buttons on the KeypadLinc are labeled like in [this button kit](https://www.smarthome.com/products/fanlinc-button-kit-for-keypadlinc-white).  When I was first setting it up with a Hub 2, I remember that the Insteon app wouldn't allow the FanLinc to be a controller of any scenes.  This basically makes the app really awkward to use.  If you change the light state / fan speed from the FanLinc dialog in the app, it actually affects the light/fan, but the keypad buttons don't update.  If instead you change the state on the from KeypadLinc dialog in the app, it changes the button states, but not the actual light/fan.  The only way I was able to get both to update when controlling from the Hub 2 was to use scenes for fan high/med/low/off, and light on/off.  Turning the light on to some arbitrary light level (and having the button state update) wasn't really possible without having a pre-made scene for each level of interest.

That all said, the value I could imagine for the scene topic on a FanLinc would be that one could set up the FanLinc as a controller of the KeypadLinc, and use the scene topic to control the light level.  Assuming that doing so would actually cause the FanLinc to trigger the KeypadLinc to update its state, then that would be a reasonably easy setup -- at least for the light.  Would there also be scene topics for the fan speeds?  If so, then it could simplify integrating things into Home Assistant if you could point HA to only the FanLinc (via the fan speed scene topics), and count on the FanLinc to update the KeypadLinc's button states to match the current fan speed.

I actually have yet to get my FanLinc / KeypadLinc integrated into Home Assistant in a way that satisfies me.  (So far, I only have the FanLinc exposed to HA, so any changes I make from HA would only change the actual light/fan state, not the keypad buttons.)  I was expecting that once I finished creating modem scenes in my USB PLM, I could use a [template fan](https://www.home-assistant.io/integrations/fan.template/) to trigger a corresponding modem scene for each desired fan speed.  (My initial attempt with a template fan was to use the scene topics for each button of the KeypadLinc, but that didn't work.  According to my notes, the KeypadLinc 'didn't seem to respond to "Trigger ALL-Link Command" (0x30) commands at all' -- at least for the non-load button groups.)

krkeegan:
Got it.

Right now, you could use the Scene topic on the keypadlinc buttons to control both of them.  So if you turn on or off the Fan Medium button on the KPL using the Scene topic for that buttob, the fan would turn to medium and the KPL medium light would turn on.  That said, you state input "the fan is on medium" would come from a different entity than you would be sending your command input to.  This can all be done in HA just fine, and isn't that hard, but you have to think it through a bit when setting it up and it probably will never be a complete auto-discovery thing no matter what we do.

We will never be able to do quite what you describe in using the FanLinc to control the KPL buttons.  For this conversation, the fan only has 1 group, the fan group.  You could use that one group to control as many other objects as you like but they all have to react to that 1 group.  There is no way you to use the scene command to turn the fan to medium and only turn on the medium kpl button if the high and low kpl buttons are also linked to the same group.

Basically, in all the scenarios I can think of, using the scene topic on a real controller (switch, dimmer, kpl, modem) is the right answer.  The only time when using the fanlinc as a controller would make any sense is if you had some weird controller-less group.  Like, two fanlincs where you wanted to turn them both to medium, but there was no linked corresponding physical controller.  Even then, I think the modem scene would be the right answer.

I am always reluctant to remove functionality, but in this case I feel like it has the potential to cause more user confusion than the benefit it provides (if any).

krkeegan:
Alright, this works well.   All scene code is now contained a in single location for Dimmer, Switch, and KPL.  Any changes will affect all devices.

I am merging this now, because of the scope of the changes I want as many people to run it before we push to master.

krkeegan:
A funny note.  I found a device that otherwise functions perfectly, but does not respond to the scene command.  It is the same model information as all others purchased in the batch.  
`DIMMABLE_LIGHTING (0x01): '2477D' (0x20) 'SwitchLinc Dimmer (Dual-Band)' firmware: 0x40`

It acks the command, but never sends a broadcast.  Tried resetting it, no change.  Links all looks the same as other functioning devices.

PR #324, opened 2021-01-12
Adding support for Two Groups and all configurable items to Hidden do…

### LINKED ISSUE #309: Support needed for Hidden Door Sensor two Groups
The hidden door sensor can be configured to report in two different ways:

1. Open/Closed on one group (Group 0x01) 
2. Open on Group 0x01 and Closed on Group 0x02

Also note: 

- Low Battery is reported on group 0x03
- Heart Beat is reported on group 0x04

Please see developers notes:  http://cache.insteon.com/developer/2845-222dev-102013-en.pdf

According to the docs, hidden door sensors are supported by Insteon-MQTT under the Battery Sensor device, however battery sensor does not support two groups.

Battery sensor supports on/off only on group 0x01, low battery on 0x02 and heartbeat on 0x04, so this does not align with the two groups group 0x02 or low bat on 0x03 for the hidden door sensor.

Note: I have a door sensor set in two groups mode.  My use case is this:  Hidden door sensor installed on basement door.  I have it linked to my basement light.  If in one group mode, it would turn on the light when I opened the door and walk into the basement, but then turn it off when I closed the door behind me.  Setting this to Two Group allows me to link it to group one to turn on the light when the door is opened, but with close reported on group 2, it takes no action on close.

tommycw1:
…or sensor (fix #309)

Signed-off-by: tommycw1 <tommycw10@gmail.com>

Fix for issue #309

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -89,9 +89,13 @@ insteon:
       - 48.3d.46
       - 48.b0.ad: 'dim1'
 
-    # Battery powered sensors (door, hidden door, window).
+    # Battery powered sensors (door, window).
     battery_sensor:
       - 94.a9.12
+
+    # Battery powered hidden door sensors
+    hidden_door:
+      - 32.a0.4b: 'back'
 
     # Battery powered motion sensors.
     motion:
@@ -448,6 +452,32 @@ mqtt:
     #   state = 'dawn'/'dusk'
     dawn_dusk_topic: 'insteon/{{address}}/dawn'
     dawn_dusk_payload: '{{is_dawn_str.upper()}}'
+
+  #------------------------------------------------------------------------
+  # Hidden Door Sensors
+  #------------------------------------------------------------------------
+
+  # Hidden Door Sensors will use the state and low battery configuration
+  # inputs from battery_sensor and add battery voltage which is configured
+  # here.
+  #
+  # To register the hidden door sensor in Home Assistant use MQTT binary
+  # sensor with a configuration like:
+  #   binary_sensor:
+  #     - platform: mqtt
+  #       state_topic: 'insteon/aa.bb.cc//battery_voltage'
+  #TODO       device_class: 'light'
+  hidden_door:
+    # Output topic and payload.  This message is sent
+    # whenever the device is polled and a new voltage, low battery voltage
+    # or heart beat interval is obtained from the device.
+    # Available variables for templating are:
+    #   address = 'aa.bb.cc'
+    #   name = 'device name'
+    #   batt_volt = raw insteon voltage level
+  
+    battery_voltage_topic: 'insteon/{{address}}/battery_voltage'
+    battery_voltage_payload: '{"voltage" : {{batt_volt}}}'
 
   #------------------------------------------------------------------------
   # Leak sensors
--- a/docs/mqtt.md
+++ b/docs/mqtt.md
@@ -30,6 +30,7 @@ be identified by it's address or the string "modem".
    - [Battery Sensors](#battery-sensors)
    - [Dimmers](#dimmers)
    - [FanLinc](#fanlinc)
+   - [Hidden Door Sensors](#hidden-door-sensors)
    - [IOLinc](#iolinc)
    - [KeypadLinc](#keypadlinc)
    - [Leak Sensors](#leak-sensors)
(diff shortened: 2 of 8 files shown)


insteon_mqtt/device/HiddenDoor.py
```diff
@@ -0,0 +1,652 @@
...
+        # requests.  Otherwise, a request may get queued multiple times
+        self._battery_request_time = 0
+
+    #-----------------------------------------------------------------------
+    @property
+    def battery_voltage_time(self):
+        """Returns the timestamp of the last battery voltage report from the
+        saved metadata
+        """
+        meta = self.db.get_meta('HDS')
```
tstabrawa, line 126: In a [comment](https://github.com/TD22057/insteon-mqtt/pull/285#issuecomment-751823375) on #285, @krkeegan indicated that he prefers a flatter metadata format.  It may be worth considering doing that for this device, since it's new code being added.  I'll leave it for you and Kevin to decide, though.

insteon_mqtt/device/HiddenDoor.py
```diff
@@ -0,0 +1,652 @@
...
+
+        Args:
+          voltage: (int) The low voltage value
+          on_done: Finished callback.  This is called when the command has
+                   completed.  Signature is: on_done(success, msg, data)
+        """
+        if voltage is not None:
+            LOG.info("Hidden Door %s cmd: set low voltage= %s", self.label,
+                     voltage)
+            self.battery_low_voltage = voltage
```
tstabrawa, line 191: It may be better to wait until we get the ack before saving the new value.

insteon_mqtt/device/HiddenDoor.py
```diff
@@ -0,0 +1,652 @@
...
+        This is called when we get a response to the
+        set_low_battery_voltage() command. Update stored low bat voltage in
+        device DB and call the on_done callback with the status.
+
+        Args:
+          msg (InpStandard): The response message from the command.
+          on_done: Finished callback.  This is called when the command has
+                   completed.  Signature is: on_done(success, msg, data)
+        """
+        self.db.set_meta('low_battery_voltage', voltage)
```
tstabrawa, line 482: This metadata value doesn't appear to be used otherwise, though, there is an HDS:battery_low_voltage in the metadata.

insteon_mqtt/device/HiddenDoor.py
```diff
@@ -0,0 +1,652 @@
...
+        # 5 minute increments for 1-255.  0 = 24 hours (default)
+        hb_interval = msg.data[5]
+        if hb_interval > 0:
+            hb_interval_minutes = hb_interval * 5
+        else:
+            hb_interval_minutes = 24 * 60 # convert 24 hours to minutes
+        LOG.info("\tHeart beat interval raw level is %s or %s minutes ",
+                 hb_interval, hb_interval_minutes)
+
+        # D7 Low Better Level
```
tstabrawa, line 445: typo ("Better")

tstabrawa (commented):
Here are a few things I noticed.  Do with it as you wish.  (... and thanks for your contribution!)

docs/mqtt.md
```diff
@@ -1060,6 +1061,121 @@ A sample motion sensor topic and payload configuration is:
...
+
+To set configuraiton option on the device, first press and hold the device link
+button until it beeps and the LED starts flashing.  Next tell insteon-mqtt the 
+device is awake with:
+
+   ```
+   { "cmd" : "awake" }
+   ```
+
+This will tell insteon-mqtto to send the command right away rather than
```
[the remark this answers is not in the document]
tommycw1: fixed

insteon_mqtt/device/HiddenDoor.py
```diff
@@ -0,0 +1,652 @@
...
+from .. import message as Msg
+
+LOG = log.get_logger()
+
+
+class HiddenDoor(BatterySensor):
+    """Insteon battery powered hidden sensor.
+
+    A hidden door sensor is basically an on/off sensor except that it's
+    battery powered and only awake when water is detected or the set button
```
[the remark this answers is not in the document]
tommycw1: fixed

insteon_mqtt/device/HiddenDoor.py
```diff
@@ -0,0 +1,652 @@
...
+    It can be configured in "one group" mode or "two group" mode:
+
+    In one group mode the open is broadcast as on ON command to group 0x01
+    and closed state is broadcast as an OFF command to group 0x01
+
+    In two group mode, open is broadcast as an ON command to group 0x01 and
+    closed is broadcast as on ON commmand to group 0x02
+
+    This was done to allow direct automoation links from this device to
+    operate dirrecent devices for open and closed.
```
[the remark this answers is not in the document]
tommycw1: fixed

insteon_mqtt/device/HiddenDoor.py
```diff
@@ -0,0 +1,652 @@
...
+          protocol (Protocol):  The Protocol object used to communicate
+                   with the Insteon network.  This is needed to allow the
+                   device to send messages to the PLM modem.
+          modem (Modem):  The Insteon modem used to find other devices.
+          address (Address): The address of the device.
+          name (str):  Nice alias name to use for the device.
+        """
+        super().__init__(protocol, modem, address, name)
+
+        self.signal_dawn = Signal()
```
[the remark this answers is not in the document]
tommycw1: Weird, thought I removed this already.  Anyway fixed.

insteon_mqtt/device/HiddenDoor.py
tommycw1: Thanks for this.  I basically copied this out of the motion sensor that already existed.  I'll let @krkeegan tell me if he wants something different.

insteon_mqtt/device/HiddenDoor.py
tommycw1: I copied this from the motion sensor doing the same thing.  @krkeegan - thoughts?

insteon_mqtt/device/HiddenDoor.py
tommycw1: fixed

insteon_mqtt/device/HiddenDoor.py
tommycw1: Yeah - reviewing this more closely I think I understand this better now.  I've removed this.

insteon_mqtt/device/HiddenDoor.py
```diff
@@ -19,8 +19,8 @@ class HiddenDoor(BatterySensor):
     """Insteon battery powered hidden sensor.
 
     A hidden door sensor is basically an on/off sensor except that it's
-    battery powered and only awake when water is detected or the set button
-    is pressed.
+    battery powered and only awake for a short time after open or closed is
+    derected or the set button is pressed.
```
tstabrawa, line 23: typo ("derected")

tommycw1: fixed

insteon_mqtt/device/HiddenDoor.py
krkeegan: Yeah, @tstabrawa is right.  I was wrong in writing that in leak and now I have to figure out a way to back that out of people's data files.

If you can just change to like 
```suggestion
        time = self.db.get_meta('HDS_battery_voltage_time')
```

Same thing on the setter, no need to wrap it in a dictionary.

I can also fix it if you like.

config.yaml
```diff
@@ -449,6 +453,41 @@ mqtt:
...
+  
+    battery_voltage_topic: 'insteon/{{address}}/battery/state'
+    battery_voltage_payload: '{"voltage" : {{batt_volt}}}'
+
+    low_battery_voltage_topic: 'insteon/{{address}}/low_battery/state'
+    low_battery_voltage_payload: '{"low_voltage_threshold" : {{low_batt_volt}}}'
+
+    heart_beat_interval_topic: 'insteon/{{address}}/heartbeat/state'
+    heart_beat_interval_payload: '{"heart_beat_interval" : {{hb_interval}}}'
+
```
krkeegan, line 490: I realize a user can change theses, but few users understand templates, and even fewer read the config file.  It would be better if these topics in their default state didn't deviate from what other devices use.  
1. There shouldn't be a "/state" on the battery topic.
2. I think the "battery" topic should work as much like it does for other devices, that is in it's default state emitting 'ON' if the battery is low and not the voltage.  So the payload would be '{{is_low_str.upper()}}'.  The additional voltage variable is nice and users can alter their config if they want it.
3. Is the 'low battery' topic an input?  Is that needed, are users likely to change their low battery voltage amount programatically?  I think this is better as a `set-flag` option unless you can convince me otherwise.  Same thing for the heartbeat.

krkeegan: Oh, I see the battery/state is an additional topic.  I still think this is probably unnecessary and adding too many topics, but if you want to add it, I could go with `insteon/{{address}}/battery_voltage`

docs/mqtt.md
```diff
@@ -1060,6 +1061,121 @@ A sample motion sensor topic and payload configuration is:
...
+Beyond these flags there are two additional settings:
+
+   - Low Battery threshold.  This is the raw Insteon voltage level that where
+   the device will trigger a group 0x03 low battery warning. Example to set to
+   64 below:
+   
+   ```
+   { "cmd" : "set_low_battery_voltage", "voltage" : 64 }
+   ```
+
```
krkeegan, line 1163: Do you have any idea about the voltage conversion to the raw values here?  At minimum it would be nice to tell the user what that calculation is.  It might also be nice to accept a nice float of a value of voltage and convert that to a raw byte for the user.

insteon_mqtt/device/HiddenDoor.py
tommycw1: It might be better if you fix this.  I'm not sure I follow completely.

docs/mqtt.md
tommycw1: The dev notes doc gives some clues.  The state these points:

61=~1.75V 
54=~1.6V
51=~1.5V
40=~1.25V (default low battery mark)

(Note these are HEX raw values)

They note that these are approx values only.  

I put together a spreadsheet and these values don't equate to something linear, so I've never bothered trying to translate from the raw number to a voltage.  To be honest I'm not sure it really matters what the voltage is.  I just want to know when the level is approaching what I typically see as my warning level so I can take action.  Whether that is presented as 40 HEX, 64 DEC or 1.25V, I don't really care.  I put the sheet here if you want to see:

https://docs.google.com/spreadsheets/d/18M4HsAcQHARZTWSjM2e51YdY5MXqat1uqpXT1fBnC8I/edit?usp=sharing

config.yaml
tommycw1: 1 & 2. OK, if you agree, I'll change these three data items to:
`insteon/{{address}}/battery_voltage`
`insteon/{{address}}/low_battery_voltage`
`insteon/{{address}}/heart_beat_interval`

I'm also open to just removing the low bat level and heart beat interval (see explanation below in 3), let me know.

3. When you send the get extended status message to the HDS, you get three data items:  Current battery voltage (raw level), Battery low level threshold (raw level), and heart beat interval (raw level).  So, since we attempt go get the battery level form the device anyway, these other two items come free.  While I suspect they wouldn't be used often, I figured why not report them over MQTT for those who want them.  So to answer your question, no this is not an input channel, this is output only.

docs/mqtt.md
krkeegan: Fair enough, can you add the few points that you have to the documentation.  Many users are not super sophisticated and this would cut down on support requests.

config.yaml
krkeegan: Got it.  I would just drop them as topics then.  It seems to me these values should just be output in the UI response to the get_extended status message.  Output, MQTT topics should be things we anticipate users building automation around.

Again, just trying to keep things consistent and clean for the average user.

tommycw1: Sounds good - I'll do that.

docs/mqtt.md
tommycw1: Thanks, I will add this to the docs.

config.yaml
tommycw1: complete

docs/mqtt.md
tommycw1: Complete.

insteon_mqtt/device/HiddenDoor.py
tommycw1: I think I have everything done other than this item.  I'll leave this one to you if that's OK.

PR #333, opened 2021-01-20
Remove Link_Cleanup_Report from Broadcast Callbacks

krkeegan:
After fixing #283, there is no longer a need to pass these messages on to the callbacks.

- Not all devices send these messages
- Some devices allow users to disable these messages
- The messages do not contain any valid ON/OFF data that the callback can process
- Removing them simplifies the callback code.
- We shouldn't expect to handle such a low level messaging at the device level

### DIFF
--- a/insteon_mqtt/handler/Broadcast.py
+++ b/insteon_mqtt/handler/Broadcast.py
@@ -28,7 +28,10 @@ class Broadcast(Base):
     Finally a broadcast LINK_CLEANUP_REPORT is sent.  This message indicates
     if the device received ACKs from all linked devices or not.  This message
     indicates that the device is finished sending messages.  However, as
-    broadcast message, it is not guaranteed to be received.
+    broadcast message, it is not guaranteed to be received.  Some devices
+    even have a user option to turn off these messages.  These messages
+    contain information about how the success or failure of the broadcast but
+    do not contain the ON/OFF value of the broadcast.
 
     This handler will call device.handle_broadcast(msg) for the device that
     sends the message.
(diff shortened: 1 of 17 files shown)


insteon_mqtt/handler/Broadcast.py
```diff
@@ -28,7 +28,10 @@ class Broadcast(Base):
     Finally a broadcast LINK_CLEANUP_REPORT is sent.  This message indicates
     if the device received ACKs from all linked devices or not.  This message
     indicates that the device is finished sending messages.  However, as
-    broadcast message, it is not guaranteed to be received.
+    broadcast message, it is not guaranteed to be received.  Some devices
+    even have a user option to turn off these messages.  These messages
+    contain information about how the success or failure of the broadcast but
```
tstabrawa, line 33: This feels like awkward wording.  Maybe the word "how" doesn't belong here?
```suggestion
    contain information about the success or failure of the broadcast but
```

tstabrawa (commented):
I only saw one possible typo (extra word in a sentence for a comment).  Other than that, and the one new flake8 error, the changes look good to me.

I also see the following tests that still test with LINK_CLEANUP_REPORT.  You might consider removing these, like you did with the other device tests.

    insteon-mqtt/tests/device$ grep LINK_CLEANUP_REPORT *
    test_DimmerDev.py:        (0x01,Msg.CmdType.LINK_CLEANUP_REPORT, 0x00, None),
    test_IOLincDev.py:        (Msg.CmdType.LINK_CLEANUP_REPORT, None),
    test_KeypadLincDev.py:        (0x01,Msg.CmdType.LINK_CLEANUP_REPORT, 0x00, None),
    test_OutletDev.py:        (0x01,Msg.CmdType.LINK_CLEANUP_REPORT, 0x00, None),
    test_RemoteDev.py:        (0x01,Msg.CmdType.LINK_CLEANUP_REPORT, 0x00, None),
    test_SmokeBridgeDev.py:        (0x01,Msg.CmdType.LINK_CLEANUP_REPORT, 0x00, None),
    test_SwitchDev.py:        (0x01,Msg.CmdType.LINK_CLEANUP_REPORT, 0x00, None),
    test_ThermostatDev.py:        (0x01,Msg.CmdType.LINK_CLEANUP_REPORT, 0x00, None),

krkeegan:
I was leaving the tests, I mean they still work just fine, they are just redundant testing things over and over again.  Not sure if that is bad form, I have never really learned the proper practices for unit tests.

PR #352, opened 2021-02-15
Refactor MQTT client configuration to ensure consistency

### LINKED ISSUE #350: MQTT topic subscriptions are not established
https://github.com/TD22057/insteon-mqtt/blob/0771befc54ba67423fca3bd6050cbfd88f3d5478/insteon_mqtt/network/Mqtt.py#L293

I noticed that my system had stopped working properly a few weeks ago, but didn't get a chance to track down the cause until today. The line referenced above was part of a commit to change the subscription behavior to only happen after `_on_connect` is called in the MQTT network link... however, on my system that function is not being invoked by the paho-mqtt client.

I've got:

paho-mqtt==1.5.1
insteon-mqtt @ git+https://github.com/td22057/insteon-mqtt@4203ec2a6c926df7d1b07d0684b07aea289e8a52 (latest 'dev' branch as of today)

I've added a `breakpoint()` as the first code line in `_on_connect`, and it's not being invoked. However, the MQTT connection *is* established, and insteon-mqtt properly publishes the states of the Insteon devices (and their state changes from local control events). It just never subscribes to the topics for the devices, so any commands I send via MQTT are ignored.

Can anyone suggest a possible reason why paho-mqtt is failing to invoke this callback?

kpfleming:
When an explicit MQTT client ID is configured, the previously-created MQTT client gets reinitialised so that it will be used. This results in the previously-configured callback functions being lost.

A new setup_client function has been added so that both code paths will configure the MQTT client the same way.

Closes #350.

### DIFF
--- a/insteon_mqtt/network/Mqtt.py
+++ b/insteon_mqtt/network/Mqtt.py
@@ -59,8 +59,19 @@ class Mqtt(Link):
         self._reconnect_dt = reconnect_dt
         self._fd = None
 
-        # Create the MQTT client and set the callbacks to our methods.
-        self.client = paho.Client(client_id=self.id, clean_session=False)
+        self.setup_client()
+
+    #-----------------------------------------------------------------------
+    def setup_client(self):
+        """ Create or reinitialise the MQTT client and set the callbacks.
+        """
+
+        client_args = {'client_id': self.id, 'clean_session': False}
+
+        if not hasattr(self, 'client'):
+            self.client = paho.Client(**client_args)
+        else:
+            self.client.reinitialise(**client_args)
         self.client.on_connect = self._on_connect
         self.client.on_disconnect = self._on_disconnect
         self.client.on_message = self._on_message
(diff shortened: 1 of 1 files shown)


krkeegan:
I confirm the initial bug from #350 occurs for me as well if I set an MQTT id.  I can also confirm that this PR fixes it.

kpfleming:
Thanks, I've cleaned up the code a bit as well since the initial push.

kpfleming:
Should be ready now, the linter and tests are happy :-)

krkeegan:
I see that.  Thank you!  That is awesome.

You can make multiple commits in the future.  It doesn't really matter here, but some people in different circumstances may get irritated when you force push things as they may have checked out your prior commits.

kpfleming:
Indeed, although in this case the first version of the code wasn't really very good, and I didn't want to leave it in the history :-)

krkeegan:
fair enough, I know the feeling.  Nice catch btw.  I don't think many people use the id setting, so this might have taken a while to find.

PR #353, opened 2021-02-15
Small Fixes to Hidden Door Sensor Code

krkeegan:
- Changes the hidden door sensor meta storage.  Only affect on users would be to make a check_battery request earlier than anticipated the first time after update.
- Updates the code to match the recent Refactored Base design.

I am merging this now as the current Hidden Door Sensor code may not function given the Refactor that just occurred.  There are no unit tests for the Hidden Door Sensor!  So I have to assume that everything works.  @tommycw1 since you have these devices can you pull the dev branch and confirm that this is working for you?  If there are issues please create an Issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ to add new devices and features. This is the reason for the jump to 0.8.0.
 
 ### Additions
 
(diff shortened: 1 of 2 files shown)


tommycw1:
I tested this with my HDS and worked as expected.  Thanks!

PR #358, opened 2021-03-01
Move Handle_Refresh with Level from ResponderBase to DimmerBase

krkeegan:
## Proposed change
Moves handle_refresh with level support from ResponderBase to DimmerBase.  This was a mistake I made in #334.  There is no affect on the end user, this is only a change that alters how the code appears to a developer.

## Checklist


tstabrawa:
Upon reviewing the changes in the PR, I can confirm that all remaining unresolved items in #334 are addressed now.  Thanks for making these changes.

Issue #351, opened 2021-02-15
Error trying to delete modem link


tstabrawa:
I tried updating one of my scenes today to remove a device from the scene, and noticed the following errors.

Immediately after running 'refresh modem' from the CLI ...

On the CLI, if I attempt to sync scenes, I see this error:

```
>>>insteon-mqtt config.yaml sync --no-refresh --run modem
Commanding Modem device 31.19.10 (modem) cmd=sync
Syncing 31.19.10 (modem) device 
  Deleting ID: 48.d8.bf (front gfci outl grp:  21 type: CTRL  data: 0x15 0x00 0x00:
ERROR: Modem db update failed: OutAllLinkUpdate: 48.d8.bf grp: 21 Cmd.DELETE ack: False
ERROR: Sync complete
```

The log shows the following when the above command is run:

```
2021-02-15 00:18:12.066 INFO Mqtt: MQTT message insteon/command/modem b'{"cmd": "sync", "refresh": false, "dry_run": false, "session": "2620907030"}'
2021-02-15 00:18:12.067 UI Mqtt: Commanding Modem device 31.19.10 (modem) cmd=sync
2021-02-15 00:18:12.067 DEBUG Mqtt: MQTT publish insteon/command/modem/session/2620907030 {"type": "MESSAGE", "data": "Commanding Modem device 31.19.10 (modem) cmd=sync"} qos=0 ret=False
2021-02-15 00:18:12.068 INFO Modem: Device 31.19.10 (modem) cmd: sync
2021-02-15 00:18:12.068 UI Modem: Syncing 31.19.10 (modem) device 
2021-02-15 00:18:12.068 DEBUG Mqtt: MQTT publish insteon/command/modem/session/2620907030 {"type": "MESSAGE", "data": "Syncing 31.19.10 (modem) device "} qos=0 ret=False
2021-02-15 00:18:12.070 DEBUG CommandSeq: CmdSeq ModemSync Running 1 of 1
2021-02-15 00:18:12.070 UI Modem:   Deleting ID: 48.d8.bf (front gfci outl grp:  21 type: CTRL  data: 0x15 0x00 0x00:
2021-02-15 00:18:12.071 DEBUG Mqtt: MQTT publish insteon/command/modem/session/2620907030 {"type": "MESSAGE", "data": "  Deleting ID: 48.d8.bf (front gfci outl grp:  21 type: CTRL  data: 0x15 0x00 0x00:"} qos=0 ret=False
[...]

PR #359, opened 2021-03-01
Add Get_Flags Function to Modem

krkeegan:
## Proposed change
Adds support for the `get_flags` command on the modem.  The value is only outputted to the log and the UI, it is not saved.

## Additional information
This may be helpful for debugging obscure modem issues, or might be useful in the future.

- This PR is related to issue: #351

## Checklist

### DIFF
--- a/insteon_mqtt/message/OutGetModemFlags.py
+++ b/insteon_mqtt/message/OutGetModemFlags.py
@@ -1,0 +1,87 @@
+#===========================================================================
+#
+# Output insteon reset the PLM modem message.
+#
+#===========================================================================
+from .Base import Base
+
+
+class OutGetModemFlags(Base):
+    """Command requesting the modem configuration
+
+    This command will retun 6 bytes:
+      - 0x02
+      - 0x73
+      - Modem Configuration Flags
+      - Spare 1
+      - Spare 2
+      - Ack/Nak
+    """
+    msg_code = 0x73
+    fixed_msg_size = 6
+
+    #-----------------------------------------------------------------------
+    @classmethod
+    def from_bytes(cls, raw):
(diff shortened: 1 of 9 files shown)


insteon_mqtt/message/OutGetModemFlags.py
```diff
@@ -0,0 +1,87 @@
...
+
+    #-----------------------------------------------------------------------
+    def __str__(self):
+        ack = ""
+        flags = ""
+        spares = ""
+        if self.is_ack is not None:
+            ack = " ack: %s" % str(self.is_ack)
+            flags = " modem flags: %s" % str(self.modem_flags)
+            spares = " spares: %s %s" % (str(self.spare1), str(self.spare2))
```
tstabrawa, line 82: It may be more useful to print flags/spares in hex (assuming the conversion to string does so in decimal by default).

tests/handler/test_ModemGetFlags.py
```diff
@@ -0,0 +1,70 @@
...
+        msg = Msg.OutGetModemFlags(is_ack=False, modem_flags=0x01, spare1=0x02,
+                                   spare2=0x03)
+        r = handler.msg_received(proto, msg)
+        assert r == Msg.UNKNOWN
+
+        # Signal Sent
+        handler.sending_message(msg)
+        assert handler._PLM_sent
+
+        #Try a message prior to sent
```
tstabrawa, line 66: This one is actually after sending ...

tstabrawa (commented):
Recognizing that these changes are already merged, here is what I noticed.  You can decide for yourself if it's worth the hassle of another PR for these.

Issue #128, opened 2019-02-28
MQTT discovery automatic configuration


djfjeff:
Home Assistant has the MQTT discovery component that is able to automatically discover devices and correctly configure them in Home Assistant : https://www.home-assistant.io/docs/mqtt/discovery/

Basically, the device itself is responsible to publish it's configuration on a specific topic through MQTT, providing details about it's configuration. Just doing that will auto-configure the device in HA and the device will be ready to use by the time you open the HA interface!

The MQTT discovery is also not completely tied specifically to Home Assistant. It seems other automation software try to use this and some device already support it too.

I think it would be great to add this to insteon-mqtt. I will try to take a look at it but I wanted to throw the idea out in case someone want to also take a look.

TD22057:
Interesting idea.  Seems like there are a few parts to think about:
1) Config input.  User needs some way to enter the config topic (is it just 1 input?).  Might also need some way to control when this happens.
2) Device supports.  Add a method to each MQTT device to send the config topic.  Questions here would be when should this occur?  Once at each start up?  Any other way seems complicated in that we'd have to keep track of all the settings for a device type and resend the config when any item in that set changes.
3) Command line support.  Probably be useful to have a command line function to trigger sending the config for all devices, all devices of a certain type, or one device.
4) Docs.  Will need to update the config.yaml file w/ examples as well as the device doc page with some explanation for how this works and when someone might want to use it.

billchurch:
I'm very interested in this as well, and would be happy to help with building this out. Is there a roadmap or anything published for this project? Would love to start contributing.

TD22057:
@billchurch You can contribute whenever you like.  We can talk design ideas here.  Start w/ a prototype.  For each device type, send an MQTT message by hand to HA until you can get it to configure the device to work properly using the entries in the IM config file.  Once we know the "right" messages to send, we can work on how and when to send them.  We can store a flag in the device DB meta data file that tells IM if HA needs to be told about the device.  Then we can add a method to each device in mqtt/*.py to send that message if needed.  When that's done, we can create something to check those flags on start up and send the messages automatically.

GreatWhiteDan:
@TD22057 @billchurch 
It is interesting that you just posted that.   I was going to start playing with your code on my own, but I am just an intermediate coder.   I at least understand the code you built, but modifying it might be difficult.    As far as this MQTT components, I do understand that.   And here is what I did last night.

[...]

PR #381, opened 2021-04-14
Skeleton of Discovery Implementation

krkeegan:
## Proposed change
This is a functioning basic implementation of the Discovery feature.  It lacks lots of documentation, and is only implemented on Dimmers.  The idea is this:

1. There are some global yaml settings for discovery such as `discovery_topic_base` and `discovery_ha_status`
2. There is a global setting for `device_info_template` that allows the user to define a device template that can be reused as a variable in the discovery templates.  @schirner had implemented something like this, and I agree this is very handy.  This is likely to be reused in every template.
3. By default each device class will look in it's subkey under the `mqtt` key in the config.yaml file for the discovery templates.  So dimmers look to `mqtt->dimmer` by default.
4. Inside this class, a `discovery_entities` key is defined.  This key __must__ have a `component` subkey and `config` key.  The `config` key __must__ include at minimum a `unique_id`.  Other HomeAssistant config settings also need to be defined.
5. The templates have a lot of variables available to them including: `name, address, name_user_case, engine, model, firmware, modem_addr, device_info_template`.  Additionally all topics defined in yaml are also available.
6. The discovery topics are automatically generated, I don't see any reason why a user would need to edit these, they do not have any effect on how HomeAssistant functions.

#### Customization
As referenced in the `min_hops` addition, users can create new discovery classes that can be used on one or more devices.  By specifying `discovery_class` as an config_extra key for the device.  The value is a reference to a key under the `mqtt` section.  This new subkey would then need to contain a `discovery_entities` key as defined above.

## Additional information

## Checklist
- [ ] Local tests pass. **Your PR cannot be merged unless tests pass**
- [ ] Documentation added/updated

Below is a sample of the published topic `homeassistant/light/4f.23.38/4f.23.38_light/config` using the config settings shown in the attached commits:
```JSON
{
  "uniq_id": "4f.23.38_light",
  "name": "dimmer",
  "cmd_t": "insteon/4f.23.38/level",
  "stat_t": "insteon/4f.23.38/state",
  "brightness": true,
  "schema": "json",
  "device": {
  "ids": "4f.23.38",
  "mf": "Insteon",
  "mdl": "DIMMABLE_LIGHTING (0x01): '2477D' (0x20) 'SwitchLinc Dimmer (Dual-Band)'",
  "sw": "69",
  "name": "dimmer",
  "via_device": "41.ee.e6"
}
}
```

### DIFF
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -432,8 +432,8 @@ mqtt:
           {
             "uniq_id": "{{address}}_light",
             "name": "{{name_user_case}}",
-            "cmd_t": "",
-            "stat_t": "",
+            "cmd_t": "{{level_topic}}",
+            "stat_t": "{{state_topic}}",
             "brightness": true,
             "schema": "json",
             "device": {{device_info_template}}
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -190,6 +190,17 @@ mqtt:
   cmd_topic: 'insteon/command'
 
 
+  # Home Assistant implements mqtt device discovery as outlined at:
+  # https://www.home-assistant.io/docs/mqtt/discovery
+  # if discover_topic_base is defined, devices (as defined in config.yaml)
+  # announce themselves to Home Assistant. Announcing occurs once
+  # upon startup of insteon-mqtt and based on request (see announce command on
+  # modem)
+  #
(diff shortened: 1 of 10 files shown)


krkeegan:
Oh, I should note, the current design publishes all discovery entities when IM starts up and republishes them anytime it see HA restart.  It seems to be working fine.

PR #384, opened 2021-04-19
Add Discovery Support to All Devices

krkeegan:
## Proposed change
This adds the discovery platform support to the devices.  ~I still need to finish the KPL, outlets, io_lincs, and ezio4.  But this is a lot of code, so I am pushing it early.  I will likely push support for the other 4 devices tomorrow.~ Done.

So far, this seems to be working reasonably well.  The biggest time suck is getting the ideal config template for HomeAssistant for each device.

The Remotes and KPL discovery entries are going to be verbose.  I struggled over the weekend thinking about ways in which the entities for each button could be duplicated from a single definition.  I was not able to come up with one that didn't make the code overly complex or the definition for the user overly awkward.

I also made some good updates to the documentation when I discover issues that users are likely to stumble into too.

## Additional information

## Checklist
- [ ] The code change is tested and works locally.
- [x] Local tests pass.

### DIFF
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -1524,6 +1524,26 @@ mqtt:
     on_off_topic: 'insteon/{{address}}/set'
     on_off_payload: '{ "cmd" : "{{value.lower()}}" }'
 
+    # Discovery Entities - Used as part of HomeAssistant MQTT Discovery
+    discovery_entities:
+      - component: 'switch'
+        config: |-
+          {
+            "uniq_id": "{{address}}_relay",
+            "name": "{{name_user_case}}",
+            "cmd_t": "{{on_off_topic}}",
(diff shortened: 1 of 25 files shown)


krkeegan:
OK support has been added for the KPL, IOLinc, Outlet, EZIO4O, and Modem Scenes.

The modem scenes are an interesting case.  It doesn't appear that HomeAssistant really has an entity that fits these items.  An entity `scene` exists, but it only allows for sending `ON` and does not envision off commands.  A `switch` is probably to closest correct option, however, it still has a state, which is not correct.  Users will have to address this when implementing scenes in their frontend.  

Discovery Entities are only generated for Modem Scenes that appear in the GroupMap on the modem.  This should limit us to only identifying scenes that the modem can actually control something.

PR #387, opened 2021-04-23
Discovery Platform

krkeegan:
## Proposed change
This adds the Discovery Platform.

In theory all of the code is complete and has been unit tested to confirm it works as intended.

I can't really test the interaction with HomeAssistant without migrating my entire installation.  I have tested small parts and it seems to work.  I will attempt to do the migration this weekend.  I will note any issues along the way and add to the documentation.

If anyone else wants to start testing this, that would be helpful.  These changes have very little interaction with the existing code, so it shouldn't break anything.  But, I would like to work out all of the issues within Discovery before merging even into dev, because changes to the config file can't be pushed to users once they set everything up.

## Additional information

## Checklist
- [x] The code change is tested and works locally.  ~(Works locally, but interaction with HA has not been tested).~ Tested
- [x] Dog food - install and run on my own HA installation
- [x] Local tests pass.
- [x] Add documentation about migrating to Discovery
- [x] Add documentation about any gotchas discovered while testing


krkeegan:
Progress:
image

krkeegan:
I am mostly up and running.  I ran into some bugs I had to fix.  But it generally seems to be working.

There are some quirks:
1. HA does not save the state between restarts, if you had turned off retain in IM, all devices will start out as off after any restart.  I am going to have to re-enable retain.
2. It isn't clear that this is the best form of configuration.  But it does seem to be the direction that HA is going, and it does add some new features (mostly cosmetic).
3. Transitioning from yaml entities to mqtt discovered entities is not easy.  At least not for me, particularly if you created a lot of unique entities to solve various issues and limitations.

There is a small issue with modem scenes.  Because of the order in which IM loads, when the discovery topics are published, we have not yet loaded the scenes file and so do not have the names of scenes.  It isn't a big deal, but users would end up with scene numbers which may be confusing.  I may have to add a signal to cause the discovery topics to publish only after we are fully loaded.

Anyways, I still have more testing and tinkering to do.  I only just got things vaguely functioning again with all of my automations.

krkeegan:
Oh, also, it does cause __a lot__ of mqtt traffic on startup.  I am publishing/subscribing to about 431 topics for 87 devices.  It takes about 5-10 seconds for all of the mqtt activity to die down after an IM restart and that is with an intel i5 processor.  Raspberry pi devices will likely struggle a bit more.

krkeegan:
OK, I am satisfied that this works as expected.  I am going to merge this into `dev` tonight along with the other waiting PRs in preparation for the next release.  I am a bit busy this weekend, so the release will likely be the middle of next week.  So please send any PRs if you see something wrong.

I am also going to recommend that people NOT go through the process of migrating to the discovery platform unless they really want to.  Until we do the fixes with the config files, migrating is a tedious process.  Even once we improve the config design, migrating will still be a good amount of work for not much benefit.

The real benefit is for new users, for them this is an unconditional improvement.

roopesh:
I'm super duper lucky that I just installed insteon-mqtt yesterday and started looking for auto discovery... how fortunate it's coming soon!

If I'm using the HA add-on, can I somehow install this branch?

krkeegan:
If you are comfortable with git, these bare instructions should get you what you need.  You can just checkout the `dev` branch.

If not, this should be pushed to release next Tuesday/Wednesday

Issue #398, opened 2021-05-14
can't send messages to FanLinc group 2


lnr0626:
**Describe the bug**
An exception is thrown when attempting to send a message to my FanLinc for group 2

**To Reproduce**
For context, I'm using the hassio add-on installation on the latest version (`0.9.0`)

1. have a fanlinc joined/paired with the modem
2. Open the Web UI
3. execute the following command:
```
>>>insteon-mqtt config.yaml on -g 2 master-fan
```
6. See error:
```
Commanding fan_linc device 47.6b.d0 (master-fan) cmd=on
ERROR: Error running command on on device 47.6b.d0 (master-fan)
```

**Expected behavior**
I expected the fan to have turned on and something similar to the following output:
```
Commanding fan_linc device 47.6b.d0 (master-fan) cmd=on
Device state updated to on=True
```

**Log Output**
Please change your logging level to 10 by editing Logging.level in your config.yaml file.

```txt
# Put your relevant logging output below this line
2021-05-13 22:48:55.715 DEBUG Mqtt: Received PUBLISH (d0, q1, r0, m256), 'insteon/command/master-fan', ...  (78 bytes)
2021-05-13 22:48:55.715 DEBUG Mqtt: Sending PUBACK (Mid: 256)
2021-05-13 22:48:55.715 INFO Mqtt: MQTT message insteon/command/master-fan b'{"cmd": "on", "level": 255, "group": 2, "reason": "", "session": "3218423953"}'
2021-05-13 22:48:55.715 UI Mqtt: Commanding fan_linc device 47.6b.d0 (master-fan) cmd=on
2021-05-13 22:48:55.715 DEBUG Mqtt: Sending PUBLISH (d0, q0, r0, m55), 'b'insteon/command/master-fan/session/3218423953'', ... (86 bytes)
2021-05-13 22:48:55.715 DEBUG Mqtt: MQTT publish insteon/command/master-fan/session/3218423953 {"type": "MESSAGE", "data": "Commanding fan_linc device 47.6b.d0 (master-fan) cmd=on"} qos=0 ret=False
2021-05-13 22:48:55.715 INFO ResponderBase: Device 47.6b.d0 grp: 2 cmd: on normal
2021-05-13 22:48:55.715 ERROR Mqtt: Error running command on on device 47.6b.d0 (master-fan)
Traceback (most recent call last):
[...]

PR #399, opened 2021-05-14
add group 2 to fanlinc responders group

lnr0626:
## Proposed change
I spent a small amount of time looking into the error I reported in #398 and found the `responder_groups`  list doesn't appear to be updated correctly for fan lincs.

This adds group 2 into the list of responder groups for fanlinc devices

## Additional information
I'm not familiar enough with the code to know if this is an appropriate change; however, it seemed like it would help move me past the exception I'm running into.

- This PR is related to issue: #398

## Checklist

Do you have examples you can point me to that demonstrate how to test this type of change?


krkeegan:
Good catch, that looks right to me.  I will test it out on my own fanlinc and merge it.

PR #406, opened 2021-05-17
fix issue with hassio initial config setup

lnr0626:
## Proposed change
This fixes the initial config setup on a new install (it current bails because `/opt/insteon-mqtt/config.yaml` does not exist)

## Checklist

### DIFF
--- a/hassio/entrypoint.sh
+++ b/hassio/entrypoint.sh
@@ -6,7 +6,7 @@
 
 if [ ! -f /config/insteon-mqtt/config.yaml ]; then
     echo "Copying default config.yaml"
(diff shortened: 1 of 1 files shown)


lnr0626:
:-/ this test is also failing on the dev branch. it looks like it may have been added in the commit to address the dns issue (https://github.com/TD22057/insteon-mqtt/pull/405/commits/a14f7d64ecb4afc073c32db827a78de3f6d7fd32). did you forget to add that config yaml to the repo @krkeegan?

krkeegan:
Oh, I never noticed this was wrong.  Thank you for fixing this.

And yes, I failed to add a test file.  Not sure how I managed to skip the CI testing.

PR #413, opened 2021-05-17
allow sending raw commands to devices

lnr0626:
## Proposed change
This arose while debugging #411 (and messing around with the other commands I found in the command tables). Essentially I found it a lot easier to debug the actual commands when I had the ability to send raw insteon commands through mqtt

I don't think this is a common use case, and there's some cleanup that needs to happen if this were desired (tests, docs, etc); but am happy to work through those if this is would be desired

## Checklist

### DIFF
--- a/insteon_mqtt/device/base/Base.py
+++ b/insteon_mqtt/device/base/Base.py
@@ -505,7 +505,8 @@ class Base:
         if data is None:
             msg = Msg.OutStandard.direct(self.addr, cmd1, cmd2)
         else:
-            msg = Msg.OutExtended.direct(self.addr, cmd1, cmd2, bytes(data), crc_type)
+            padded_and_trimmed_data = (data + ([0] * 14))[:14]
+            msg = Msg.OutExtended.direct(self.addr, cmd1, cmd2, bytes(padded_and_trimmed_data), crc_type)
 
         if ext_resp:
             # Use the standard command handler which will notify us when the
--- a/insteon_mqtt/device/base/Base.py
+++ b/insteon_mqtt/device/base/Base.py
@@ -509,13 +509,13 @@ class Base:
             msg = Msg.OutExtended.direct(self.addr, cmd1, cmd2, bytes(padded_and_trimmed_data), crc_type)
 
         if ext_resp:
-            # Use the standard command handler which will notify us when the
-            # command is ACK'ed.
-            msg_handler = handler.StandardCmd(msg, self.handle_std_raw_message, on_done)
-        else:
             # Use the extended response command handler which will notify us when the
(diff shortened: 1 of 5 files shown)


krkeegan:
I can understand how something like this would be helpful.

I think you should allow a full extended message to be defined, not just all zero bytes.  That gets a bit unweildy from the command line, but from a json mqtt message, the extended bytes can just be a list.

lnr0626:
It does allow full extended commands - the mqtt message takes a data list, and the cli takes a `-d`/`--data` argument that's zero padded and trimmed to 14 bytes - the `--ext-req` flag is purely a convenience shorthand :-)

krkeegan:
Ahh, I messed that. That is perfect then.

Yes, this would have been quite helpful at times.  I had failed to realize it could be added so easily.

insteon_mqtt/device/base/Base.py
```diff
@@ -493,6 +494,31 @@ def set_flags(self, on_done, **kwargs):
...
+        This can be used to send commands that aren't supported directly
+        by insteon-mqtt yet. This is as low level as it gets, use at your
+        own risk
+        """
+        if data is None:
+            msg = Msg.OutStandard.direct(self.addr, cmd1, cmd2)
+        else:
+            msg = Msg.OutExtended.direct(self.addr, cmd1, cmd2, bytes(data), crc_type)
+
+        if ext_resp:
```
lnr0626, line 512: Haha whoops, this condition is backwards

insteon_mqtt/device/base/Base.py
```diff
@@ -1149,6 +1175,16 @@ def handle_flags(self, msg, on_done):
                "{:08b}".format(msg.cmd2))
         on_done(True, "Operation complete", msg.cmd2)
 
+    #-----------------------------------------------------------------------
+    def handle_std_raw_message(self, msg, on_done):
```
lnr0626, line 1179: I'll probably just combine these handlers, I want sure if the response/signature was different based on standard/extended messages

lnr0626:
Cool beans, I'll add some docs and tests to get this ready to merge.

I'll also move the pad/trim logic to the message handler, and may switch the cmd name to `raw_command` or `raw` - do you have a preference?

krkeegan:
I don't have a name preference.  I can see a value in either.

krkeegan:
Looks good to me, unless you have any more changes I will merge.

lnr0626:
I think it's good to go, feel free to merge away :-)

krkeegan:
This works in my setup. I didn't test extensively, but a basic test of standard commands worked as expected.  Thank you for this, it is very helpful.

PR #414, opened 2021-05-18
Create Base Config; Overlay User Config on Top of Base Config

### LINKED ISSUE #383: Create a Base Config.yaml and Apply User Config.yaml Over the Top
**Is your feature request related to a problem? Please describe.**
I think this has bubbled up before.  But while working on Discovery this became more apparent to me.  Taking the Discovery changes as an example.  When users update, if they want to enable the feature they will have three options (really only two).

1. Adopt `config-example.yaml` as their new config file and insert their insteon devices and config settings into this new file.
2. Keep their own `config.yaml` file and copy and past portions of `config-example.yaml` into their config.
3. Read `config-example.yaml` and write everything on their own (this isn't really how this happens, I don't think).

The issue, is that when we make changes to the default config, unless a user is paying attention they likely don't receive the benefit of these changes.  Basically, we leave the merging of their config file with the new config settings up to the user.

The obvious upside, is that we are unlikely to affirmatively break a user's installation.

**Describe the solution you'd like**
I was wonder whether we could do something I commonly see in linux config files.  That is have a base config file which the user is instructed not to change.  This file resides inside the application directory so when the application is updated this file is also updated with our changes.

A user can then change the values of these default settings by redefining the settings in their own config file.  So if we change a default setting and the user is using their own customized setting, they would not see the change.

**Describe alternatives you've considered**
I can't think of another good alternative other than status quo.

**Additional context**
So this is a little complex.  Deep dict merging is not supported below python 3.9 (it has some new feature).  So we would have to implement our own function.  This doesn't sound too bad (famous last words).  But we would have to pay attention to how certain merges are done, but I think our config file is relatively simple so I think it can be done reasonable?

**Thoughts?**

krkeegan:
## Breaking change
There should not be any breaking changes.  But this is a dramatic shift in how the config file is parsed, so we should use caution.

## Proposed change
This creates a config-base.yaml file that is contained in the package.  The user defined config file loaded from the commandline is overlayed or merged over the top of the config-base.yaml file.  This means that settings not defined in the user config file, will use the settings from the config-base.yaml file.

The concept, is that the user should not edit the config-base.yaml file.  Instead, they should overwrite any settings using their user config file.

This means, going forward, if we make changes to a payload template for example, the change will be pushed to all users who have not defined a custom payload template.  I suspect that very few users will define many of the templates themselves.

However, users with config files prior to this version, will have user config files that contain all of the default settings.  These users will continue to use the settings defined in their user config files.

This makes switching to the Discovery Platform much easier for existing users.  It also cuts down the size of the default user yaml file to ~180 lines from the prior ~1800 length.

## Additional information
- This PR fixes or closes issue: closes #383

## Checklist
- [x] Local tests pass.
- [x] I would like to run this for a few days before merging 

After merging I would like to wait a bit before pushing to master


krkeegan:
Hmm, not sure why `tests/network/test_Hub.py` line 106 failed initially.  It works fine on my computer, and passed when I re-ran the test suite. Will have to keep an eye on this, perhaps we have some race condition happening in our testing.

krkeegan:
Alright, I am content that this works well without issue.  I have paired back my own config file to just the basics without issue.  I am merging this into dev if anyone else wants to test it out.

Otherwise, I will merge it this weekend into master.

PR #417, opened 2021-05-19
Convert Default State Payloads to JSON; Add Attributes by Default

krkeegan:
## Breaking change
If merged as part of 1.0.0 nothing will break.  Users with config files prior to this will have settings that override these changes.

## Proposed change
Changes the default payload of the state topics and a few others to use JSON payloads with additional attributes such as `reason`, `mode`, `timestamp`, and any others that are available.

Also updates the default discovery entities for HomeAssistant so that these attributes are exposed to HomeAssistant by default.

Basically, this exposes more of the functionality in InsteonMQTT to the user by default.

## Additional information
This should be merged with 1.0.0.

## Checklist
- [x] Local tests pass.
- [x] Tests have been updated to verify that the new code works.

### DIFF
--- a/insteon_mqtt/data/config-base.yaml
+++ b/insteon_mqtt/data/config-base.yaml
@@ -282,8 +282,8 @@ mqtt:
                      {%- endif -%}",
             "cmd_t": "{{scene_topic}}",
             "device": {{device_info_template}},
-            "payload_on": "{\"cmd\": \"on\", \"group\": \"{{scene}}\"}",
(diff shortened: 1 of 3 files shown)


krkeegan:
I have been running this for about 5 days with only a handful of devices using custom discovery classes.  The majority of my setup is now using the default configuration settings.  Everything seems to be working as expected and I am not seeing any errors in HA or in IM.    I feel comfortable saying that this doesn't have any huge flaws.

PR #423, opened 2021-05-25
Resume dim for non i2cs devices

### LINKED ISSUE #415: Enable Resume Dim Feature on non-I2CS Devices
**Is your feature request related to a problem? Please describe.**
The Resume Dim feature on #411 only works on i2cs devices.  Non-i2cs devices need the command to be sent as a standard length message.

**Describe the solution you'd like**
Add logic to send set-flags command to non-i2cs devices using standard length messages.

lnr0626:
## Proposed change

Enables the resume_dim flag for non i2cs devices

## Additional information

- This PR fixes or closes issue: fixes #415 

## Checklist

I only have i2cs devices, so will need help verifying this works properly for non i2cs devices :-)

### DIFF
--- a/insteon_mqtt/device/base/Base.py
+++ b/insteon_mqtt/device/base/Base.py
@@ -664,11 +664,11 @@ class Base:
                       resume dim)
         """
         cmd1 = Msg.CmdType.SET_OPERATING_FLAGS
-        if self.db.engine == 2:
-            return Msg.OutExtended.direct(self.addr, cmd1, cmd2,
-                                            bytes([0x00] * 14), crc_type='D14')
-        else:
+        if self.db.engine is not None and self.db.engine < 2:
             return Msg.OutStandard.direct(self.addr, cmd1, cmd2)
+        else:
+            return Msg.OutExtended.direct(self.addr, cmd1, cmd2,
+                                          bytes([0x00] * 14), crc_type='D14')
 
     #-----------------------------------------------------------------------
     def sync(self, dry_run=True, refresh=True, sequence=None, on_done=None):
--- a/insteon_mqtt/device/base/Base.py
+++ b/insteon_mqtt/device/base/Base.py
@@ -588,6 +588,10 @@ class Base:
         if self.db.desc is None or self.db.firmware is None or force:
             seq.add(self.get_model)
 
+        # If the engine is not known, or force true, run get_engine
+        # if self.db.engine is None or force:
+        #     seq.add(self.get_engine)
+
     #-----------------------------------------------------------------------
     def get_flags(self, on_done=None):
         """Get the Insteon operational flags field from the device.
--- a/insteon_mqtt/device/base/Base.py
+++ b/insteon_mqtt/device/base/Base.py
@@ -589,8 +589,8 @@ class Base:
             seq.add(self.get_model)
 
         # If the engine is not known, or force true, run get_engine
-        # if self.db.engine is None or force:
-        #     seq.add(self.get_engine)
+        if self.db.engine is None or force:
+            seq.add(self.get_engine)
 
     #-----------------------------------------------------------------------
     def get_flags(self, on_done=None):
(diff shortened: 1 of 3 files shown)


krkeegan:
Awesome, I appreciate you making that a separate method so that it can be reused for other flags.

krkeegan:
My only hesitation here is the addition of the `get-engine` request to the `refresh` request.  I see it is limited to only occur in cases where the engine is unknown or refresh is called with `force`.

I understand the benefit of this, making things "just work" for users.  The concern that I have always had, is as we chain more commands to `refresh` we run a greater risk of complications.

I think this is probably a good addition, but I want to think about it for a minute.

lnr0626:
that makes sense - I added it mainly because i had a few devices that didn't have that set initially, and without that the devices likely wouldn't respond correctly. 

An alternative would be to add some additional handling so that if we don't know the engine and we received a NACK (invalid checksum) in response to a set_operating_flags message, we either update the engine to i2cs and retry; or run `get_engine` and then retry

my main hesitation with just setting the engine to i2cs is that's a fairly large assumption that may break at some point.

that might also have the potential for infinite loops if (for some reason), we are never able to set the engine?

tests/device/test_DimmerDev.py
```diff
@@ -73,11 +73,23 @@ def test_handle_on_off_manual(self, test_device, group, cmd1, cmd2, expected):
...
+    def test_set_resume_dim_non_i2cs(self, test_device):
         # set_resume_dim(self, resume_dim, on_done=None)
         for params in ([True, 0x04], [False, 0x05]):
             test_device.set_resume_dim(resume_dim=params[0])
             assert len(test_device.protocol.sent) == 1
+            assert isinstance(test_device.protocol.sent[0].msg, Msg.OutStandard)
+            assert test_device.protocol.sent[0].msg.cmd1 == 0x20
+            assert test_device.protocol.sent[0].msg.cmd2 == params[1]
+            test_device.protocol.clear()
+
```
krkeegan, line 85:
```suggestion

    def test_set_resume_dim_unknown(self, test_device):
        # set_resume_dim(self, resume_dim, on_done=None)
        test_device.db.engine = None
        for params in ([True, 0x04], [False, 0x05]):
            test_device.set_resume_dim(resume_dim=params[0])
            assert len(test_device.protocol.sent) == 1
            assert isinstance(test_device.protocol.sent[0].msg, Msg.OutStandard)
            assert test_device.protocol.sent[0].msg.cmd1 == 0x20
            assert test_device.protocol.sent[0].msg.cmd2 == params[1]
            test_device.protocol.clear()
            
```

insteon_mqtt/device/base/Base.py
```diff
@@ -649,6 +653,23 @@ def get_model(self, on_done=None):
...
+        Args:
+          cmd2 (int): the set operating flags option (i.e. 0x04 to enable
+                      resume dim)
+        """
+        cmd1 = Msg.CmdType.SET_OPERATING_FLAGS
+        if self.db.engine == 2:
+            return Msg.OutExtended.direct(self.addr, cmd1, cmd2,
+                                            bytes([0x00] * 14), crc_type='D14')
+        else:
+            return Msg.OutStandard.direct(self.addr, cmd1, cmd2)
```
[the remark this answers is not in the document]
krkeegan: It probably won't matter since we are going to add `get_engine` to `refresh` but given that all devices made in the last 5+ years are i2cs, we should default to that and not the old standard length message.

tests/device/test_DimmerDev.py
krkeegan: Adding a test for if the engine is unknown.  Again, this is probably overkill.

krkeegan (changes requested):
I tested out `refresh` with the `get_engine` command on a few i2cs devices and some old i1 devices and it all works fine.  We already do a model request, so it likely makes sense to get the engine version too.

A few small additions, but I am otherwise happy to merge this.

lnr0626:
those changed looked good to me, so just went ahead and commited them. thanks for the reviews and the help with testing this stuff out!

krkeegan (approved):

PR #441, opened 2021-09-09
Add fan state to fan_linc discovery info

tstabrawa:
## Proposed change

This change adds state_topic and state_value_template to the discovery information for the fan component of fanlinc devices.  Without these topics, Home Assistant treats the fan as always off.

## Additional information

- This PR fixes a problem where Home Assistant displays the state of fanlinc fans as "off" even though they are actually on.

### Without fix:
image
image

### With fix:
image
image

## Checklist

  - Tested by overriding discovery_entities in config.yaml and applying as custom discovery class.
  - Four tests fail both on master and on my branch locally.  I assume this is just a problem with my environment.
  - No new code added
  - N/A

  - N/A


krkeegan:
It seemed like something was off in the fan.  Thank you for doing this.

PR #484, opened 2022-03-02
MQTT DIscovery features: named entities, device and class level overrides of entity attributes

kpfleming:
## Proposed change
This set of changes allows users to override, and add, attributes supplied by the default MQTT Discovery templates without having to copy-and-paste them (which avoids the need to update those copies when the standard versions are modified). In addition users can completely suppress entities that would be supplied by the default template.

The applications for this feature are numerous:
* Changing component type ('switch' -> 'light')
* Changing HA name pattern
* Adding HA attributes ('icon', 'suggested_area')
* Suppressing unwanted entities (especially for KeypadLinc and remotes)

## Checklist

- [ ] The code change is tested and works locally.
- [ ] Local tests pass. **Your PR cannot be merged unless tests pass**
- [ ] Code documentation was added where necessary

- [ ] Documentation added/updated


TD22057:
Kevin,
Thanks for working on this.  Just an FYI but you don't have to push every individual commit (I have a an inbox full of messages saying "1 commit pushed")  It might be better to wait until you have the feature done or big blocks of it done and you want review to push those commits up to this PR.  Thanks - Ted

kpfleming:
Ahh sorry, I've got the PR in draft mode so that it's clear that it isn't ready for review yet, and I'm using this branch as a deployment tool to get the code onto my machines, so I have to push everything as I get it ready for internal testing :-)

Feel free to unsubscribe, and I can ping you when I move the PR out of 'draft' mode (the commits will have been rebased/squashed by then as well).

kpfleming:
@TD22057 @krkeegan This is ready for review. Sorry for the notification noise while it was under development :-)

krkeegan:
I need to sit down this weekend and test this out on my installation.

kpfleming:
I tried to avoid creating any breaking changes; support for the `config` item as a string in discovery classes is still supported, just not documented. There's even a test case for it at line 122 in https://github.com/TD22057/insteon-mqtt/blob/dev/tests/configs/discovery_schema_good.yaml.

PR #531, opened 2024-11-11
If not all DB records are received, try one-by-one

tstabrawa:
Also, improve DeviceDbGet & DeviceRefresh test coverage.

## Proposed change

Thanks for considering this PR!

This PR changes the DeviceRefresh handling so that, after attempting a bulk database download (with DeviceDbGet), if the database is incomplete, Insteon-MQTT will attempt to retrieve the missing DB entries one-by-one.

This feature was previously proposed by @krkeegan in [this message](https://github.com/TD22057/insteon-mqtt/issues/233#issuecomment-742697299) on #233:

> I thought about this a bit more. I am still of the mindset that the current system using the single request is preferable. However, I thought of a way to add in support for individual requests as a fallback. The process would look like:
> 
> 1. Perform the db get as we do currently. (at the start of this the cached db is emptied)
> 2. Whenever a) the process finishes OR b) the process times out
> 3. Check the cached DB
> 4. If the cached DB lacks either a) a last_entry entry OR b) is missing entries for addresses (the addresses are non-contiguous)
> 5. Then spawn a separate handler to a) specifically query the missing entries AND b) query entries at the end until the last_entry is found.
> 
> Basically, the separate handler is patching up the holes of what was missed.
> 
> The two downsides to this are:
> 1. Some complexity in the code (but keeping this as a separate handler should contain things)
> 2. We will lose the "future proof" feature of not needing to know the address layout in the devices.  This is a bummer, but probably not a big deal.

The one-by-one downloading is performed in DeviceScanManagerI2, to reduce the effect of the added complexity on the rest of the DeviceRefresh/DebiceDbGet code.

Notably, this one-by-one fallback approach [appears to be the same technique employed by pyinsteon](https://github.com/pyinsteon/pyinsteon/blob/9e719ac97ff003cb963cd0da0533face1ff847b4/pyinsteon/aldb/aldb.py#L76) (which is used by Home Assistant's native Insteon integration).

Flake8 passes without errors, pylint shows no new warnings, and all new code is shown as covered by pytest coverage report.  Also, DeviceDbGet and DeviceRefresh are now listed as 100% covered.

## Additional information

- This PR fixes or closes issue: fixes #
- This PR is related to issue: #233
- Link to documentation pull request: N/A

Example force-refresh output for a particularly persnickety device is attached (device ID's have been changed to so as not to dox innocent devices):
- [db-2nd-pass.txt](https://github.com/user-attachments/files/17695641/db-2nd-pass.txt)
- [db-2nd-pass2.txt](https://github.com/user-attachments/files/17695645/db-2nd-pass2.txt)

## Checklist

- [x] Documentation added/updated -- N/A

### DIFF
--- a/insteon_mqtt/db/Device.py
+++ b/insteon_mqtt/db/Device.py
@@ -175,6 +175,23 @@ class Device:
         return delta == self.delta
 
     #-----------------------------------------------------------------------
+    def is_complete(self):
+        """See if the database is complete.
+
+        Check that all DB entries between START_MEM_LOC and self.last
+        (inclusive) are present in either self.entries or self.unused.
+
+        Returns:
+          (bool) Returns True if the database is complete.
+        """
+        if (self.last not in self.entries.values() and
+                self.last not in self.unused.values()):
+            # Last entry hasn't been added/downloaded yet
+            return False
+        expected_entries = ((START_MEM_LOC - self.last.mem_loc) / 8) + 1
+        return len(self.entries) + len(self.unused) == expected_entries
+
+    #-----------------------------------------------------------------------
     def increment_delta(self):
         """Increments the current database delta by 1
 
@@ -770,7 +787,7 @@ class Device:
     #-----------------------------------------------------------------------
     def _add_using_new(self, addr, group, is_controller, data,
                        on_done):
-        """Add a anew entry at the end of the database.
+        """Add a new entry at the end of the database.
 
         First we send the new entry to the remote device.  If that works,
         then we update the previously last entry to mart it as "not the last
--- a/insteon_mqtt/db/DeviceScanManagerI2.py
+++ b/insteon_mqtt/db/DeviceScanManagerI2.py
@@ -1,0 +1,121 @@
+#===========================================================================
+#
+# Device Scan Manager for i2 or newer Devices
+#
+#===========================================================================
+from .. import log
+from .. import message as Msg
+from .. import util
+from .. import handler
+from .DeviceEntry import DeviceEntry
+from .Device import START_MEM_LOC
+
+LOG = log.get_logger()
+
+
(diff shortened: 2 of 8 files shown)


krkeegan:
Thank you for doing this.

I realize I have been absent for a while.  I need to spend some time getting back up to speed on how everything works.  Hopefully I can get to this by next week.  We also have a number of other changes that need to be made.  Some changes in HomeAssistant require some modifications on our end.

tstabrawa:
No rush from my end.  I actually started on these changes over a year ago, and then life got busy.

On the other hand, if you were looking for an excuse to dive back in, then you're welcome.  :-)

krkeegan:
This seems to be working for me just fine.  At least normal refreshing is.  I can't simulate a missing entry, but I gather from your use that this is working.

I will let this run on my local instance until Sunday and then push a release then if everything keeps going well.

tstabrawa:
> This seems to be working for me just fine. At least normal refreshing is. I can't simulate a missing entry, but I gather from your use that this is working.

FWIW, the tests added in tests/db/test_DeviceScanManagerI2.py (`test_handle_record`) are intended to simulate various scenarios of missing entries.  It's not quite a full sequence, but the idea was to cover all of the corner cases I could come up with.  For a full sequence, feel free to review the debug traces attached in my initial message on this PR:
- [db-2nd-pass.txt](https://github.com/user-attachments/files/17695641/db-2nd-pass.txt)
- [db-2nd-pass2.txt](https://github.com/user-attachments/files/17695645/db-2nd-pass2.txt)

Also, FWIW, tests/handler/test_DeviceRefresh.py (`test_dbget_finish`) has some steps towards the end for detecting and responding to an incomplete initial DB download.

> I will let this run on my local instance until Sunday and then push a release then if everything keeps going well.

Thanks!

PR #537, opened 2024-11-12
Allow Installation of Insteon_MQTT Module via Pip in Addon

### LINKED ISSUE #536: Alpine and Python Now Restrict Installing Packages Through PIP
This was done because using PIP can break an OS installation of python.  This means no one can install the addon currently in HomeAssistant.

This isn't a concern for us inside our own docker container, luckily there is a command line argument to get around this.

krkeegan:
Fixes #536

Adds a command line argument to suppress the pip error.


krkeegan:
This is confirmed working.

PR #539, opened 2024-11-14
Enable Optional Encryption Without CA_Cert

### LINKED ISSUE #535: CA_Cert is Not Required for TLS Connection to MQTT
A CA_Cert is required if you are running your own certificate authority, but if you are using a commonly accepted certificate authority such as Let'sEncrypt, the certificate is already trusted by the OS.

The only bummer is I used ca_cert as a test to determine if Encryption should be enabled.  So, I have to fix that in the process.

krkeegan:
## Proposed change
Adds the following enable setting to `config.yaml`
```
mqtt:
  encryption:
    enable: False
```

Previously encryption was enabled by setting the `ca_cert` value.  But this is not necessary if your servers cert is from a trusted authority such as Let's Encrypt.

I added a deprecation warning to the documentation, but for the time being if `ca_cert` is set encryption will continue to be enabled even if `enable: False`.  This will only affect users who have enabled encryption previously and now attempt to disable it with `encryption: False`.  Since I suspect basically no one updates their config files with settings that are added in later, I doubt this will ever be an issue.

Also encryption seems to be a little used feature.

## Additional information
- This PR fixes or closes issue: fixes #535 

## Checklist
- [x] Local tests pass. 

### DIFF
--- a/config-example.yaml
+++ b/config-example.yaml
@@ -170,6 +170,9 @@ mqtt:
   #retain: 1
 
   encryption:
+    # Whether to enable a TLS encrypted connection to your MQTT broker.
+    enable: False
+
     # Encryption Options for encrypted broker connections
     # These settings will be passed to the `tls_set()` method.  Please refer
     # to the Paho client documentation for details:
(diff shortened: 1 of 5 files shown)


tstabrawa (approved):

krkeegan:
This is also working.

PR #540, opened 2024-11-14
Remove Battery Devices from *_All Commands

### LINKED ISSUE #533: Battery Sensor Devices Should not Be Included in *_All Commands
This applies to sync_all, refresh_all, join_all and any others

I think our initial thought had been that these messages would be cached for battery devices, but that doesn't seem to work out correctly.  I have to revisit why I concluded this isn't working, but the solution was just to skip battery devices in the all commands.

krkeegan:
## Proposed change
I remember why this was an issue and have updated issue #533.

In short battery devices are present, the sequence for the *_all command will halt when it gets to a battery device and only resume when that device has woken up and the command was sent.

This is at best confusing to the user, because they expect to see the sequence run to completion and receive a message like
"Refresh All Devices Complete"

At worst, this is problematic if a device after a battery device in the sequence is not getting the proper command and the user
assumes that it is.

_Note_
There is an alternative solution in which the commands are sent immediately to battery devices with the assumption that they are asleep and the command will be queued.  This seems bad to me as the devices may be awake, in which case they could collide with each other.  Also the messages (other than `refresh` for which there is a test) would just pile up in the battery device's queue potentially with numerous duplicates if these were called a lot.

Overall, I think they best option is for users to call these commands on each desired battery device directly.

## Additional information
- This PR fixes or closes issue: fixes #533

## Checklist
- [x] Local tests pass.

### DIFF
--- a/insteon_mqtt/Modem.py
+++ b/insteon_mqtt/Modem.py
@@ -437,6 +437,9 @@ class Modem:
         the database sizes.  So it usually should only be called if no other
         activity is expected on the network.
 
+        Battery devices are not included in this command and this command
+        must be run individually on each battery device.
+
         Args:
           force (bool):  Force flag passed to devices.  If True, devices
                 will refresh their Insteon db's even if they think the db
@@ -469,6 +472,9 @@ class Modem:
         lose your data directory.  Otherwise you likely never need to use
         this.
 
+        Battery devices are not included in this command and this command
+        must be run individually on each battery device.
+
         Args:
           on_done:  Finished callback.  This is called when the command has
                     completed.  Signature is: on_done(success, msg, data)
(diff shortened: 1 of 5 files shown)


insteon_mqtt/Modem.py
```diff
@@ -478,7 +486,8 @@ def get_engine_all(self, on_done=None):
 
         # Reload all the device databases.
```
tstabrawa, line 487: I might be worth it to update this comment while we're in here.  Looks like it was copy-pastad from refresh-all.

```suggestion
        # Run Get Engine on all the devices
```

tstabrawa (approved):

krkeegan:
This is also working fine on my local instance.

PR #541, opened 2024-11-14
Migrate HomeAssistant Config Files Location

### LINKED ISSUE #532: Move HomeAssistant Config Files Location
Config files for addons are not supposed to be stored in /config anymore

See:
https://developers.home-assistant.io/blog/2023/11/06/public-addon-config/
https://developers.home-assistant.io/docs/add-ons/configuration/

We need to:

- [ ] Modify the addon to comply with the new standards
- [ ] Move user's current configs to the new path. Not sure how to do this, perhaps look how other addons handled this change.

krkeegan:
## Proposed change
The new location will be backed up by a Home Assistant snapshot and is Home Assistant's preferred location for storing config files going forward.

1. If a `config.yaml` file is __not__ found in the new location; AND
2. A `config.yaml` file is found at the old location; THEN
3. Copy the entire contents of the old config directory to the new one
4. Add a notice file to the old config directory telling users:
- The new location of their config files
- That they can delete these config files if everything is working correctly.

This process should ensure that no data is lost and should provide enough guidance to users on where their config files are located.

## Additional information
- This PR fixes or closes issue: fixes #532 

## Checklist
- [x] Local tests pass.

I have updated and added comments about the location of the new config files in every location in documentation that I can think of.

### DIFF
--- a/hassio/entrypoint.sh
+++ b/hassio/entrypoint.sh
@@ -6,14 +6,15 @@
 # to the add-on configuration folder.
 # We add a file to the old directory so that hopefully a user
 # doesn't get confused in the future.
-if [ ! -f /config/config.yaml ] && [ -f /homeassistant/insteon-mqtt/config.yaml]; then
-    shopt -s dotglob
-    mv /homeassistant/insteon-mqtt/* /config/ \
-        || echo "Failed to migrate InsteonMQTT configuration"; exit 1;
+
+
+if [ ! -f /config/config.yaml ] && [ -f /homeassistant/insteon-mqtt/config.yaml ]; then
+    mv /homeassistant/insteon-mqtt/* /config/ || \
+        { echo "Failed to migrate InsteonMQTT configuration"; exit 1; }
     # If user was using default data and log locations, edit them to match the new location
-    sed -i "s|storage: '/config/insteon-mqtt/data'|storage: '/config/data'|" /config/config.yaml
+    sed -i "s|storage: '\{0,1\}\"\{0,1\}/config/insteon-mqtt/data'\{0,1\}\"\{0,1\}|storage: /config/data|" /config/config.yaml
     sed -i "s|file: /config/insteon-mqtt/insteon_mqtt.log|file: /config/insteon_mqtt.log|" /config/config.yaml
-    cp /opt/insteon-mqtt/hassio/CONFIG-MOVED.txt /homeassistant/insteon-mqtt/CONFIG-MOVED.txt
+    /bin/cp /opt/insteon-mqtt/hassio/CONFIG-MOVED.txt /homeassistant/insteon-mqtt/CONFIG-MOVED.txt
     sed -i "s/<<MIGRATION_DATE>>/$(date)/g" /homeassistant/insteon-mqtt/CONFIG-MOVED.txt
     echo "InsteonMQTT configuration successfully migrated from /config/insteon-mqtt."
 fi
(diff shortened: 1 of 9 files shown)


hassio/entrypoint.sh
```diff
@@ -1,17 +1,29 @@
...
     echo "Welcome to InsteonMQTT!"
     echo "Creating your initial config.yaml file."
-    /bin/cp /config/insteon-mqtt/config.yaml.default /config/insteon-mqtt/config.yaml
-    sed -i "s/#storage: 'data'/storage: '\/config\/insteon-mqtt\/data'/" /config/insteon-mqtt/config.yaml
-    sed -i "s/#file: \/var\/log\/insteon_mqtt.log/file: \/config\/insteon-mqtt\/insteon_mqtt.log/" /config/insteon-mqtt/config.yaml
-    echo "Please edit the file /config/insteon-mqtt/config.yaml"
+    /bin/cp /config/config.yaml.default /config/config.yaml \
+      || echo "Unable to create initial InsteonMQTT config.yaml file"; exit 1;
+    sed -i "s/#storage: 'data'/storage: '\/config\/insteon-mqtt\/data'/" /config/config.yaml
+    sed -i "s/#file: \/var\/log\/insteon_mqtt.log/file: \/config\/insteon_mqtt.log/" /config/config.yaml
```
tstabrawa, line 24: With the contents of the old `/config/insteon-mqtt` directory moved to `/config`, wouldn't that mean for legacy users, the data directory would now be in `/config/data`?  If so, wouldn't it be easier to support user errors/questions if new users have their data directories created in `/config/data` as well?

Also, sed will let you use an alternate character for the delimiter if you don't want to have to escape all of those forward-slashes.  For example, something like the following might be easier to read/maintain:

```suggestion
    sed -i "s|#storage: 'data'|storage: '/config/data'|" /config/config.yaml
    sed -i "s|#file: /var/log/insteon_mqtt.log|file: /config/insteon_mqtt.log|" /config/config.yaml
```

hassio/entrypoint.sh
```diff
@@ -1,17 +1,29 @@
...
+    echo "Please define the required settings in the file /config/config.yaml"
```
tstabrawa, line 25: Something like the following may help add-on users (who don't RTFM) find the correct file:

```suggestion
    echo "Please define the required settings in the file /config/config.yaml"
    echo "(which can be found at '/addons_configs/83fc19e1_insteon-mqtt/config.yaml'"
    echo "within VSCode or SSH Addon)."
```

hassio/entrypoint.sh
krkeegan: Good catch.

The data directory is also copies over, not sure why there would be a need to use the old directory?

tstabrawa: Keeping the contents of the old `data` directory could save people the effort of refreshing the DB for all of their devices, no?

krkeegan: Oh, I see what you are saying.

1. For new users, the data directory will be /config/data
2. For migrated users
- We move the entire contents of insteon-mqtt to config (this includes moving data)
- I think this is best, it seems like HomeAssistant may not allow access to the entire homeassistant_config folder for much longer
- Plus if we move the data to the addon config, it will get backed up with the addon

But you are right, we need to modify the users config to point to the new location.  I think it is best to do that automatically.  I doubt many, if any, have moved this location.

krkeegan: OK, I pushed another commit that edits the location of the storage files and the logfile for users as long as they are using the default location.

tstabrawa (approved):

krkeegan:
I had to fix a few things, but the migration script works now!

There will be errors if a user has used a non-default log location or storage directory.  This will need to be heavily noted in the release.

PR #542, opened 2025-01-10
Add MQTT Discovery Entities for Config Buttons in Home Assistant

krkeegan:
## Proposed change
This adds a couple of buttons (join, refresh, pair, and sync) to Home Assistant:

image

These buttons significantly improve the user experience when adding a device to Home Assistant.

The Log Sensor is not great, but is better than nothing.

These are only changes to the base config yaml file.

## Checklist


krkeegan:
This all works when tested locally as well.

